PR #2, opened 2015-03-11
return boolean from block command

semafor:


dbusmock/templates/urfkill.py
```diff
@@ -75,10 +75,14 @@ def block(self, index, should_block):
...
-        obj = dbusmock.get_object(path)
-        self.internal_states[objname] = new_block_state
-        obj.Set('org.freedesktop.URfkill.Killswitch', 'state', new_block_state)
-        obj.EmitSignal('org.freedesktop.URfkill.Killswitch', 'StateChanged', '', [])
+        try:
+            obj = dbusmock.get_object(path)
+            self.internal_states[objname] = new_block_state
+            obj.Set('org.freedesktop.URfkill.Killswitch', 'state', new_block_state)
+            obj.EmitSignal('org.freedesktop.URfkill.Killswitch', 'StateChanged', '', [])
+        except:
```
martinpitt: bare except: statements are utterly wrong. If there's an unexpected error in the calls, then it should be propagated as usual, and not papered over with a generic except:. That will also hide really bad stuff like SyntaxError or real programming errors. Please add a comment what the return value should mean, but I suppose it should just tell you whether the previous state was blocked or not? Determining that should work by looking at properties, not through exceptions.

Otherwise, please add a comment what could go wrong here, and only catch the expected exceptions.

tests/test_urfkill.py
```diff
@@ -73,11 +73,19 @@ def test_block(self):
         property_object = bus.get_object('org.freedesktop.URfkill', '/org/freedesktop/URfkill/WLAN')
         propiface = dbus.Interface(property_object, 'org.freedesktop.DBus.Properties')
         self.assertEqual(propiface.Get('org.freedesktop.URfkill.Killswitch', 'state'), 0)
-        iface.Block(1, True)
+
+        ret_block = iface.Block(1, True)
+        self.assertTrue(ret_block)
```
martinpitt: the variable is redundant, please just do

     self.assertTrue(iface.Block(...))

tests/test_urfkill.py
```diff
@@ -73,11 +73,19 @@ def test_block(self):
...
         self.assertEqual(propiface.Get('org.freedesktop.URfkill.Killswitch', 'state'), 1)
-        iface.Block(1, False)
+
+        ret_unblock = iface.Block(1, False)
+        self.assertTrue(ret_unblock)
         self.assertEqual(propiface.Get('org.freedesktop.URfkill.Killswitch', 'state'), 0)
 
+        # 99 is an unknown type to the mock, so it should return false.
+        ret_failed_block = iface.Block(99, False)
+        self.assertFalse(ret_failed_block)
```
martinpitt: That illustrates the above, and is very confusing. "99" is the same type as "1" (int), so maybe you mean "interface number"? Does the real urfkill really return False there, instead of raising a D-Bus error (which would be a much better and common design). Can you please check urfkill's signature for that method?

martinpitt:
Thanks for working on this! This requires some cleanup, but the overall direction is fine.

dbusmock/templates/urfkill.py
semafor: Then I'd opt not to catch anything. My bad.

tests/test_urfkill.py
semafor: Yes, that was intentional and I agree. Scary

semafor:
Pushed some fixes to your comments, Martin. Thanks for looking at this PR.

martinpitt:
Thanks for the cleanup! Merged.

PR #6, opened 2015-03-31
More NetworkManager template functionality

semafor:
Implements the following methods:

 * `org.freedesktop.NetworkManager.DeactivateConnection`
 * `org.freedesktop.NetworkManager.Settings.AddConnection`
 * `org.freedesktop.NetworkManager.Settings.Connection.Update`
 * `org.freedesktop.NetworkManager.Settings.Connection.Delete`

It also allows Connections with `autoconnect`, added using AddConnection, to be automatically connected by the first found device, roughly like NetworkManager itself does.


martinpitt:
Thank you! I rebased this to fold the third commit into the first one, fixed the dozen PEP-8 errors and added NEWS entries. I'll release 0.15 shortly (after pulling in the other request).

PR #9, opened 2015-05-01
Set RsnFlags to have the same value as WpaFlags

pete-woods:
WpaFlags really needs to have the same value as RsnFlags, or this important information is conflicting


martinpitt:
Thanks! Pulled.

PR #10, opened 2015-05-29
test_ofono: Don't check "private" fields

iainlane:
Ubuntu's version of the list_modems script obfuscates 'private' data.
Check for some other properties we set in the mock that aren't private
instead.


martinpitt:
Thanks! I'll merge this and then commit --amend / push -f a NEWS entry on top of it.

PR #11, opened 2015-09-15
Add missing properties to ethernet device and active connection

pete-woods:
Could this also be landed in the stable overlay PPA when merged? Thanks!


martinpitt:
Thanks!

PR #12, opened 2015-09-30
Origin/powerd

semafor:
Add a simple powerd mock that changes it's sysState and does wakeups.

### DIFF
--- a/dbusmock/templates/powerd.py
+++ b/dbusmock/templates/powerd.py
@@ -1,0 +1,174 @@
+'''powerd D-BUS mock template'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+"""node /com/canonical/powerd {
+  interface com.canonical.powerd {
+    methods:
+      requestSysState(in  s name, in  i state, out s cookie);
+      clearSysState(in  s cookie);
+      requestWakeup(in  s name, in  t time, out s cookie);
+      enableProximityHandling(in  s name);
+      disableProximityHandling(in  s name);
+      clearWakeup(in  s cookie);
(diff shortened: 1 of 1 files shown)


dbusmock/templates/powerd.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='s', out_signature='')
+def disableProximityHandling(self, name):
+    pass
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='s', out_signature='')
+def clearWakeup(self, cookie):
+    del self._wakeup_requests[cookie]
```
semafor: Need to kill the timer as well, here.

dbusmock/templates/powerd.py
```diff
@@ -0,0 +1,199 @@
...
+        name, cookie, time
+    ))
+    self._wakeup_requests[cookie] = (name, time)
+    t = threading.Timer(
+        time - int(timelib.time()), wakeup, args=(name, cookie, time)
+    )
+    try:
+        t.start()
+    except (KeyboardInterrupt, SystemExit):
+        print("Cancelling foo")
```
semafor: Will remove.

martinpitt:
Apparently this is still work in progress; but TBH powerd is very specific to Ubuntu Touch, so I'm not sure if it's a good idea to ship this upstream. How many projects are going to need this?

dbusmock/templates/powerd.py
```diff
@@ -0,0 +1,199 @@
...
+        t.cancel()
+
+    return cookie
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='s', out_signature='')
+def enableProximityHandling(self, name):
+    pass
+
```
martinpitt: Please don't fully spell out these methods if they don't do anything, or only do something trivial. Please rather use `AddMethods()`  and add them wholesale, like the other templates do. This should get rid of most explicitly written methods here.

semafor:
I didn't know powerd was specific to Ubuntu Touch, but if it is,  you're right. Closing for now.

PR #13, opened 2015-10-06
Rework activate method to not depend on "specific object" path

pete-woods:


martinpitt:
I did that a bit more succinctly in commit 971862e9. Thanks!

PR #14, opened 2015-10-06
Added x-canonical-private-synchronous to notifications capabilites

xavi-garcia-mena:
Added x-canonical-private-synchronous to the notifications mock capabilities, as the sound indicator needs it


martinpitt:
This capability is a private one, so it would be wrong for the mock to offer it by default in the upstream package. But you don't need to use the default, as the comment in the file says you can pass the `capabilities` parameter to the mock on instantiation.

So I don't merge this, sorry.

PR #15, opened 2015-10-07
Allow the passing of template parameters via the command-line as JSON

pete-woods:
This seems to work. Comments on my Python style welcome :)


martinpitt:
Looks good, thanks! The `print` commands don't work for python3 (see test suite failures), and the error messages should go to stderr. I fixed that up and merged.

PR #16, opened 2015-10-19
Fix indexing bug in SettingsAddConnection

pete-woods:


martinpitt:
Thanks! Pulled, but added NEWS entry and prefixed commit with "NetworkManager template:". (Please do that in the future)

PR #17, opened 2015-12-04
3 fixes for network manager template

pete-woods:


martinpitt:
This looks good, thank you! Merged with a trivial PEP-8 fix (spotted by the tests).

PR #25, opened 2017-01-20
Add NetworkManager ObjectManager support

lkundrak:
It's in the server for quite some time and now the client library relies on it.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -84,6 +84,7 @@ class DBusMockObject(dbus.service.Object):
         self.path = path
         self.interface = interface
         self.is_object_manager = is_object_manager
+        self.object_manager = None
 
         self._template = None
         self._template_parameters = None
@@ -115,6 +116,7 @@ class DBusMockObject(dbus.service.Object):
                        'GetManagedObjects', '', 'a{oa{sa{sv}}}',
                        'ret = {dbus.ObjectPath(k): objects[k].props ' +
                        '  for k in objects.keys() if ' + cond + '}')
+        self.object_manager = self
 
     def _reset(self, props):
         # interface -> name -> value
@@ -232,6 +234,7 @@ class DBusMockObject(dbus.service.Object):
                              properties)
         # make sure created objects inherit the log file stream
         obj.logfile = self.logfile
+        obj.object_manager = self.object_manager
         obj.is_logfile_owner = False
         obj.AddMethods(interface, methods)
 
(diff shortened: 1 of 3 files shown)


dbusmock/mockobject.py
```diff
@@ -232,6 +234,7 @@ def AddObject(self, path, interface, properties, methods):
                              properties)
         # make sure created objects inherit the log file stream
         obj.logfile = self.logfile
+        obj.object_manager = self.object_manager
```
martinpitt: I don't understand that bit -- `AddObject()` is not about inheritance or similar, it's a "meta-method" on the mock to create new mocked objects. These should not automatically be assumed to be object managers?

dbusmock/mockobject.py
```diff
@@ -531,6 +534,18 @@ def MethodCalled(self, name, args):
...
+            self.object_manager.EmitSignal(OBJECT_MANAGER_IFACE, 'InterfacesAdded',
+                                           'oa{sa{sv}}', [dbus.ObjectPath(path),
+                                                          objects[path].props])
+
+    def object_manager_emit_removed(self, path):
+        if self.object_manager is not None:
+            self.object_manager.EmitSignal(OBJECT_MANAGER_IFACE, 'InterfacesRemoved',
+                                           'oas', [dbus.ObjectPath(path),
+                                                   objects[path].props])
+
```
martinpitt: Can you please add some simple unit test for these to tests/test_api.py?

dbusmock/templates/networkmanager.py
```diff
@@ -26,8 +26,9 @@
 
 
 BUS_NAME = 'org.freedesktop.NetworkManager'
-MAIN_OBJ = '/org/freedesktop/NetworkManager'
-MAIN_IFACE = 'org.freedesktop.NetworkManager'
+MAIN_OBJ = '/org/freedesktop'
```
martinpitt: That looks wrong, this is almost certainly not a valid object path. It would also break backwards compatibility?

dbusmock/mockobject.py
lkundrak: That property is to track the object manager that manages this object.

dbusmock/mockobject.py
lkundrak: I'll try to.

dbusmock/templates/networkmanager.py
lkundrak: I'm not sure about this. The object manager's path is /org/freedesktop because it needs to manage the manager object (/o/fd/NetworkManager) and thus needs to be above it.

I'm not familiar with dbusmock enough to know what properties are required to be kept stable for backwards compatibility. What change here is problematic, is it the change of MAIN_OBJ property?

martinpitt: Oh, so `/org/freedesktop` is an actually valid path in NM, I just saw in d-feet. Sorry about that then. This certainly looks very strange, I had expected the main object /o/fd/NM to be the object manager too.

MAIN_OBJ is the first/main object that dbusmock creates and the only object that will remain after you call `Reset()`. While `Reset()` is API, using the latter on a template certainly doesn't make much sense, so I think we can live with that change wrt. backwards compatibility.

So I guess that part is okay then.

martinpitt:
As NetworkManager ≥ 1.6 depends on this, it is already covered by tests, so I merged it with some changelog stylistic adjustments. Thanks!

PR #30, opened 2017-06-27
Fixes bug in a TestCase causing infinite dbus stop wait

femdom:
Found that timeout never happens because the timeout is not decremented. 

Also, it would be nice to control timeout value.


martinpitt:
Thanks for spotting this! I [fixed this a little differently](https://github.com/martinpitt/python-dbusmock/commit/5f3dbe42d2e778fbf37038cef5c58bf26ac05b41) as I think a 50s timeout is too much (I intended 5s), and there is no real reason to make this configurable as that timeout shouldn't depend on what the actual test does.

PR #33, opened 2017-10-31
tests/test_timedated.py: update for v235

xnox:
v235 has changed the text output of the timedated command, upgrade
test suite to handle those strings / assertions.

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>


martinpitt:
Thanks! I fixed it a bit more generically.

PR #34, opened 2017-11-07
Travis

martinpitt:


martinpitt:
Works as intended, I'll fix the pycodestyle issues separately.

PR #42, opened 2018-03-10
NetworkManager template: Add GetConnectionByUuid method

aruiz:


martinpitt:
There are some pyflakes and pycodestyle failures: https://travis-ci.org/martinpitt/python-dbusmock/jobs/351566146

Also, bionic fails due to some package uninstallability, but that's not your fault.

tests/test_networkmanager.py
```diff
@@ -514,6 +514,27 @@ def test_settings_secrets(self):
...
+        connectionPath = self.settings.AddConnection(settings)
+
+        self.assertEqual(self.settings.GetConnectionByUuid(uuid), connectionPath)
+
+        fakeuuid = '123123123213213'
+        pathnotpresent = False
+        try:
+            self.settings.GetConnectionByUuid(fakeuuid)
+        except dbus.exceptions.DBusException as e:
+            assert("uuid" in e.args[0])
```
martinpitt: please use `self.assertIn("uuid", ..)` to get meaningful failures.

aruiz:
I think this fixes everything

martinpitt:
Still one pycodestyle error: https://travis-ci.org/martinpitt/python-dbusmock/jobs/351788018

tests/test_networkmanager.py
```diff
@@ -514,6 +514,25 @@ def test_settings_secrets(self):
...
+                'uuid': uuid,
+                'type': '802-11-wireless'}, signature='sv'),
+            '802-11-wireless': dbus.Dictionary({
+                'ssid': dbus.ByteArray('The_SSID'.encode('UTF-8'))}, signature='sv')
+        }, signature='sa{sv}')
+        connectionPath = self.settings.AddConnection(settings)
+        self.assertEqual(self.settings.GetConnectionByUuid(uuid), connectionPath)
+
+        fakeuuid = '123123123213213'
+        try:
```
martinpitt: This should use `assertRaises` (see https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises), as otherwise the test would also (wrongly) succceed if calling `GetConnectionByUuid()` with an invalid UID succeeds.

aruiz: I'm happy to use assertRaises, but note that this won't succeed in any case where the uuuid is not what I provided to make it raise, I check that the message output is exactly what I want it to be, any other exception arg would fail.

aruiz: I have to use assertRaaisesRegexp which is the only function that works on both 2.x and 3, but it is deprecated in Python 3...

tests/test_networkmanager.py
```diff
@@ -514,6 +514,23 @@ def test_settings_secrets(self):
...
+                'type': '802-11-wireless'}, signature='sv'),
+            '802-11-wireless': dbus.Dictionary({
+                'ssid': dbus.ByteArray('The_SSID'.encode('UTF-8'))}, signature='sv')
+        }, signature='sa{sv}')
+        connectionPath = self.settings.AddConnection(settings)
+        self.assertEqual(self.settings.GetConnectionByUuid(uuid), connectionPath)
+
+        fakeuuid = '123123123213213'
+        with self.assertRaisesRegexp(dbus.exceptions.DBusException, ".*uuid.*%s$" % fakeuuid):
+            self.settings.GetConnectionByUuid(fakeuuid)
```
martinpitt: Note you can use assertRaisesExp with `... as ex:` and then make assertions about the exception content, as you did before. That was actually useful :-)

Using a deprecated function in the tests is okay for now. If/once it goes away in Py 3, it can easily be aliased.

aruiz:
I believe I've solved all the issues? The only failing test now has to do with a broken package on the image, nothing to do with the code.

PR #43, opened 2018-03-28
Add support for suspend-then-hibernate

superm1:
Suspend to hibernate is a new feature in systemd, merged in:
github.com/systemd/systemd/pull/8274

It was renamed to suspend then hibernate in:
https://github.com/systemd/systemd/pull/8606


martinpitt:
Great, all green now. Do you have more stuff coming that should go into a release? Otherwise I'll do one in the next days, there's some other nice pending stuff.

superm1:
Thanks!  I don't know of anything else that needs to go in from my side.

PR #44, opened 2018-04-05
Add helpers for application testing

benzea:
I am currently working on adding testing to g-c-c and would also like to streamline the code for g-s-d. For this, I worked on the two classes in these commits, and I was wondering whether it might be feasible to merge at least the X11 helper into dbusmock in some form.


benzea:
Example of using both:

```
import os
import sys
import unittest
import dbusmock

BUILDDIR = os.environ.get('BUILDDIR', os.path.join(os.path.dirname(__file__)))

class PanelTestCase(dbusmock.X11SessionTestCase, dbusmock.GTest):
    g_test_exe = os.path.join(BUILDDIR, 'test-network-panel')

if __name__ == '__main__':
    # avoid writing to stderr
    unittest.main(testRunner=unittest.TextTestRunner(stream=sys.stdout, verbosity=2))
```

Which runs Xvfb, and the system + session busses to execute each of the tests in the binary separately.

dbusmock/gtest.py
```diff
@@ -0,0 +1,108 @@
...
+                print('')
+                print(e)
+                print('Error generating separate test funcs, will call binary once.')
+                result.test_all = result._gtest_all
+
+        return result
+
+    @staticmethod
+    def make_tests(exe, result):
+        test = subprocess.Popen([exe, '-l'], stdout=subprocess.PIPE, stderr=PIPE_DEVNULL)
```
hadess, line 49: We ran into the problem today that this particular execution is done without an X11 server started, which means that it can fail if the process needs X11 to initialise.

PR #45, opened 2018-05-08
add PairDevice 'class_' parameter

bjoernh:
The Change make it possible to define a different blutooth class. New parameter has a default value from the old hard coded bluetooth class id, so it won't break old code.

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -187,8 +187,8 @@ def AddDevice(self, adapter_device_name, device_address, alias):
 
 
 @dbus.service.method(BLUEZ_MOCK_IFACE,
-                     in_signature='ss', out_signature='')
-def PairDevice(self, adapter_device_name, device_address):
+                     in_signature='ssi', out_signature='')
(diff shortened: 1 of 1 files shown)


dbusmock/templates/bluez5.py
```diff
@@ -187,8 +187,8 @@ def AddDevice(self, adapter_device_name, device_address, alias):
 
 
 @dbus.service.method(BLUEZ_MOCK_IFACE,
-                     in_signature='ss', out_signature='')
-def PairDevice(self, adapter_device_name, device_address):
+                     in_signature='ssi', out_signature='')
+def PairDevice(self, adapter_device_name, device_address, class_=5898764):
```
martinpitt: This is not actually a backwards-compatible change - D-Bus does not support default parameters. But `PairDevice()` is not an official API of bluez anyway (that would be [Pair()](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/device-api.txt)), and we don't even properly cover this in tests as it's too flaky. So I guess this is okay then.

PR #53, opened 2019-10-22
Add GitHub action for running PR tests

martinpitt:


martinpitt:
No need to run the same tests in two CIs. This was mostly for experimentation, let's keep this on stand-by if/when we need it.

PR #54, opened 2019-11-14
Add mock server for low-memory-monitor

hadess:
API version 2.0

https://gitlab.freedesktop.org/hadess/low-memory-monitor/

### DIFF
--- a/dbusmock/templates/low_memory_monitor.py
+++ b/dbusmock/templates/low_memory_monitor.py
@@ -1,0 +1,38 @@
+'''low-memory-monitor mock template
+
+This creates the expected methods and properties of the main
+org.freedesktop.LowMemoryMonitor object.
+
+This provides only the 2.0 D-Bus API of low-memory-monitor.
+'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Bastien Nocera'
+__email__ = 'hadess@hadess.net'
+__copyright__ = '(c) 2019, Red Hat Inc.'
+__license__ = 'LGPL 3+'
+
+import dbus
+
+from dbusmock import MOCK_IFACE
+
+BUS_NAME = 'org.freedesktop.LowMemoryMonitor'
+MAIN_OBJ = '/org/freedesktop/LowMemoryMonitor'
+MAIN_IFACE = 'org.freedesktop.LowMemoryMonitor'
+SYSTEM_BUS = True
+
+
+def load(mock, parameters):
+    # Loaded!
+    mock.loaded = True
+
+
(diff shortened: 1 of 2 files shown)


hadess:
I've figured out how to run the tests, so I fix the template and the test :)

martinpitt:
Huh, travis, where are  you.. It seems this never ran in Travis, perhaps some GitHub webhook hiccup?

Anyway, some questions: This is a rather simple interface, and the test basically just tests dbusmock itself -- it doesn't call any external CLI or other "real" tool to make sure that the mock conforms to what a real client program expects. What e. g. tests/test_upower.py does against the `upower` cli, or tests/test_logind.py against `loginctl`. But I suppose that's ok, this doesn't look like the kind of service that has a standalone CLI.

But I wonder how useful this is actually going to be with shipping in dbusmock? You can send a signal with [a single method call](https://github.com/martinpitt/python-dbusmock/blob/master/tests/test_api.py#L396) to dbusmock in any test suite that uses the monitor, so it's not really simpler (actually, it's rather *harder*) to use this new template.

So, if you want to go ahead with this and wait until it made its way into all stable distros etc., I won't veto it. But I wonder if that's really an useful approach for what you plan to do? I. e. will you have that many new project tests that share this template, and avoiding the repetition of a single `.EmitSignal()` call is worth that?

Thanks!

hadess:
It allows me to focus on the client side of the problem when writing tests, as in those:
https://gitlab.gnome.org/GNOME/glib/merge_requests/1005

hadess:
Making use of GMemoryMonitor in the test isn't out of the question when it's finally landed in GLib...

martinpitt:
OK, thanks. I fixed the travis problem, can you please force-push again to trigger them? I'll land afterwards.

martinpitt:
Thanks! Some code style errors:

```
======================================================================

FAIL: test_codestyle (test_code.StaticCodeTests)

----------------------------------------------------------------------

  File "/tmp/source/tests/test_code.py", line 41, in test_codestyle

    self.assertEqual(pep8.returncode, 0, out)

AssertionError: 1 != 0 : ./tests/test_low_memory_monitor.py:27:1: E302 expected 2 blank lines, found 1

======================================================================

FAIL: test_pyflakes (test_code.StaticCodeTests)

----------------------------------------------------------------------

  File "/tmp/source/tests/test_code.py", line 34, in test_pyflakes

    self.assertEqual(flakes.returncode, 0, out)

AssertionError: 1 != 0 : ./tests/test_low_memory_monitor.py:23: 'gi.repository.GLib' imported but unused

./dbusmock/templates/low_memory_monitor.py:22: 'dbusmock.mockobject' imported but unused

./dbusmock/templates/low_memory_monitor.py:23: 'dbusmock' imported but unused
```

martinpitt:
One more to go:

   ./dbusmock/templates/low_memory_monitor.py:22: 'dbusmock.mockobject' imported but unused

hadess:
> One more to go:
> 
> ./dbusmock/templates/low_memory_monitor.py:22: 'dbusmock.mockobject' imported but unused

Sorry, I really thought I got them all otherwise I would have installed pyflakes.

PR #56, opened 2019-11-30
NetworkManager template: Add Active Connection ID

oleksandrkozlov:
The ID property was added to an active connection. This fix adds the possibility to get an active connection by ID.

### DIFF
--- a/dbusmock/templates/networkmanager.py
+++ b/dbusmock/templates/networkmanager.py
@@ -577,6 +577,7 @@ def AddActiveConnection(self, devices, connection_device, specific_object, name,
     settings = conn_obj.settings
     conn_uuid = settings['connection']['uuid']
     conn_type = settings['connection']['type']
+    conn_id = settings['connection']['id']
 
     device_objects = [dbus.ObjectPath(dev) for dev in devices]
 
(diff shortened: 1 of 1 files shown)


oleksandrkozlov:
@martinpitt Did the master fail because of my commit?

martinpitt:
@oleksandrkozlov : Tests were okay here, but indeed [this master run](https://travis-ci.org/martinpitt/python-dbusmock/jobs/619330026) repeatedly fails on rawhide, I retried it twice. So either this is flaky, and the test run here just was really lucky, or rawhide got some new NetworkManager (or depending library) in the meantime, and this is a coincidence. This needs to be looked at more deeply, surely before the next p-dbusmock release.

oleksandrkozlov:
I see. Thank you.

martinpitt:
I had a closer look at the regression at last. It's not triggered by this change, it's just a change/regression in rawhide's NetworkManager. Curiously, on the D-Bus interface the device has the expected state:
```
$ busctl introspect org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/Devices/mock_WiFi1
org.freedesktop.NetworkManager.Device
.State                                         property  u              30
```

`30` [means](https://developer.gnome.org/NetworkManager/unstable/nm-dbus-types.html#NMDeviceState) `DISCONNECTED`, so I think the mock is ok. But `nmcli` claims otherwise:

```
$ nmcli dev
DEVICE  TYPE  STATE      CONNECTION 
wlan0   wifi  connected  -- 
```

martinpitt:
I filed https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/313 about this.

PR #57, opened 2019-12-22
NetworkManager template: set "StateReason" for SetDeviceActive()/SetD…

thom311:
…eviceDisconnected()

We need to set the "StateReason". See commit 776b131e2b50 ('NetworkManager
template: add "StateReason" property to devices (#55)') for the reasons.

See-also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/ec39498fc6d70cf6887f08b4dfbc5cff021fc231
See-also: https://github.com/martinpitt/python-dbusmock/pull/55
See-also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/313


martinpitt:
Whoa, thanks @thom311 !

PR #62, opened 2020-04-30
NetworkManager template: AddConnectionUnsaved

oleksandrkozlov:
[AddConnectionUnsave](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Settings.html#gdbus-method-org-freedesktop-NetworkManager-Settings.AddConnectionUnsaved) method has been added to [org.freedesktop.NetworkManager.Settings](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Settings.html). The method has the same signature, and the same behavior as [AddConnection](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Settings.html#gdbus-method-org-freedesktop-NetworkManager-Settings.AddConnection) except the first one doesn't save a key-file on the disk, but for mocking, it doesn't matter. Therefore, the previously implemented `SettingsAddConnection` has been utilized.


oleksandrkozlov:
Hi, Martin (@martinpitt), could you please review?

oleksandrkozlov:
> Hi, Martin (@martinpitt), could you please review?

Just a kind reminder.

oleksandrkozlov:
> Thanks! I suppose that's simple enough to be without a test case then.

Thank you too. Yeah, I didn't see much sense to write tests for that.

PR #65, opened 2021-01-31
RFC: Extend BlueZ5 template

ukBaz:
I've added some additional functionality to the BlueZ5 template that I have found useful for my testing.
 - Add support for BLE beacons
 - Device discovery
 - Add device with more properties
 - Connect method for device to allow for discovery of GATT services
 - Add GATT services, characteristics

I don't think this is ready for merging but I did want to submit it for comment. Is this going in a direction that would be useful to the project?  What changes would need to be made for it to be accepted?

It has a large dictionary in the file with information on various devices. I did want this to be a separate file which would be project specific. The data was captured from GetManagedObjects during a session and has formed the basis of the test data. Feels like it shouldn't be part of the library. Any thoughts on how to have this in a separate file?

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -16,7 +16,11 @@ This supports BlueZ 5 only.
 # of the license.
 
 __author__ = 'Philip Withnall'
+__email__ = 'philip.withnall@collabora.co.uk'
 __copyright__ = '(c) 2013 Collabora Ltd.'
+__license__ = 'LGPL 3+'
+
+import time
 
 import dbus
 
@@ -33,10 +37,20 @@ PROFILE_MANAGER_IFACE = 'org.bluez.ProfileManager1'
 ADAPTER_IFACE = 'org.bluez.Adapter1'
 MEDIA_IFACE = 'org.bluez.Media1'
 NETWORK_SERVER_IFACE = 'org.bluez.Network1'
+LEDADVERTISING_MNGR_IFACE = 'org.bluez.GattManager1'
 DEVICE_IFACE = 'org.bluez.Device1'
+GATT_MNGR_IFACE = 'org.bluez.GattManager1'
+GATT_SRVC_IFACE = 'org.bluez.GattService1'
+GATT_CHRC_IFACE = 'org.bluez.GattCharacteristic1'
+GATT_DSCR_IFACE = 'org.bluez.GattDescriptor1'
+
+# here = Path(__file__).parent
+# device_json = Path.joinpath('device_db.json')
+# with device_json.open() as input_db:
+#     microbit_data = json.load(input_db)
 
 
-def load(mock, _parameters):
+def load(mock, parameters):
     mock.AddObject('/org/bluez', AGENT_MANAGER_IFACE, {}, [
         ('RegisterAgent', 'os', '', ''),
         ('RequestDefaultAgent', 'o', '', ''),
@@ -76,7 +90,7 @@ def AddAdapter(self, device_name, system_name):
             # Audio/Video Remote Control Profile (target)
             '0000110c-0000-1000-8000-00805f9b34fb',
         ], variant_level=1),
-        'Discoverable': dbus.Boolean(True, variant_level=1),
+        'Discoverable': dbus.Boolean(False, variant_level=1),
         'Discovering': dbus.Boolean(True, variant_level=1),
         'Pairable': dbus.Boolean(True, variant_level=1),
         'Powered': dbus.Boolean(True, variant_level=1),
@@ -92,6 +106,9 @@ def AddAdapter(self, device_name, system_name):
         'PairableTimeout': dbus.UInt32(180, variant_level=1),
     }
 
+    if device_name == 'hci1':
+        adapter_properties['Address'] = 'AA:01:02:03:04:05'
+
     self.AddObject(path,
                    ADAPTER_IFACE,
                    # Properties
@@ -99,11 +116,12 @@ def AddAdapter(self, device_name, system_name):
                    # Methods
                    [
                        ('RemoveDevice', 'o', '', ''),
-                       ('StartDiscovery', '', '', ''),
+                       ('StartDiscovery', '', '', DeviceDiscovery),
                        ('StopDiscovery', '', '', ''),
+                       ('SetDiscoveryFilter', 'a{sv}', '', '')
                    ])
-
     adapter = mockobject.objects[path]
+
     adapter.AddMethods(MEDIA_IFACE, [
         ('RegisterEndpoint', 'oa{sv}', '', ''),
         ('UnregisterEndpoint', 'o', '', ''),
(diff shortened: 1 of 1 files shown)


dbusmock/templates/bluez5.py
```diff
@@ -16,7 +16,11 @@
 # of the license.
 
 __author__ = 'Philip Withnall'
+__email__ = 'philip.withnall@collabora.co.uk'
```
hadess, line 19: Is this required?

dbusmock/templates/bluez5.py
```diff
@@ -16,7 +16,11 @@
...
 __copyright__ = '(c) 2013 Collabora Ltd.'
+__license__ = 'LGPL 3+'
```
hadess, line 21: Or this? (seeing as the whole library is under that license)

dbusmock/templates/bluez5.py
```diff
@@ -33,10 +37,20 @@
...
+GATT_DSCR_IFACE = 'org.bluez.GattDescriptor1'
+
+# here = Path(__file__).parent
+# device_json = Path.joinpath('device_db.json')
+# with device_json.open() as input_db:
+#     microbit_data = json.load(input_db)
 
 
-def load(mock, _parameters):
+def load(mock, parameters):
```
hadess, line 53: Is there any reason to rename that variable? Seems unused.

dbusmock/templates/bluez5.py
```diff
@@ -76,7 +90,7 @@ def AddAdapter(self, device_name, system_name):
             # Audio/Video Remote Control Profile (target)
             '0000110c-0000-1000-8000-00805f9b34fb',
         ], variant_level=1),
-        'Discoverable': dbus.Boolean(True, variant_level=1),
+        'Discoverable': dbus.Boolean(False, variant_level=1),
```
hadess, line 93: This is a good idea, but I would also change the `Discovering` variable just below, and do that in a separate commit, with an explanation that this is the way BlueZ operates nowadays.

dbusmock/templates/bluez5.py
```diff
@@ -33,10 +37,20 @@
...
 MEDIA_IFACE = 'org.bluez.Media1'
 NETWORK_SERVER_IFACE = 'org.bluez.Network1'
+LEDADVERTISING_MNGR_IFACE = 'org.bluez.GattManager1'
 DEVICE_IFACE = 'org.bluez.Device1'
+GATT_MNGR_IFACE = 'org.bluez.GattManager1'
+GATT_SRVC_IFACE = 'org.bluez.GattService1'
+GATT_CHRC_IFACE = 'org.bluez.GattCharacteristic1'
+GATT_DSCR_IFACE = 'org.bluez.GattDescriptor1'
+
+# here = Path(__file__).parent
```
hadess, line 47: Remove this debug.

dbusmock/templates/bluez5.py
```diff
@@ -92,18 +106,22 @@ def AddAdapter(self, device_name, system_name):
         'PairableTimeout': dbus.UInt32(180, variant_level=1),
     }
 
+    if device_name == 'hci1':
+        adapter_properties['Address'] = 'AA:01:02:03:04:05'
```
hadess, line 110: It would be better to create a separate function that "calculated" an address and use that just above. Please do that in a separate commit.

dbusmock/templates/bluez5.py
```diff
@@ -92,18 +106,22 @@ def AddAdapter(self, device_name, system_name):
...
                    [
                        ('RemoveDevice', 'o', '', ''),
-                       ('StartDiscovery', '', '', ''),
+                       ('StartDiscovery', '', '', DeviceDiscovery),
                        ('StopDiscovery', '', '', ''),
+                       ('SetDiscoveryFilter', 'a{sv}', '', '')
                    ])
-
     adapter = mockobject.objects[path]
+
```
hadess, line 124: Really no reason to add a new line here.

dbusmock/templates/bluez5.py
```diff
@@ -92,18 +106,22 @@ def AddAdapter(self, device_name, system_name):
...
     self.AddObject(path,
                    ADAPTER_IFACE,
                    # Properties
                    adapter_properties,
                    # Methods
                    [
                        ('RemoveDevice', 'o', '', ''),
-                       ('StartDiscovery', '', '', ''),
+                       ('StartDiscovery', '', '', DeviceDiscovery),
```
hadess, line 119: As mentioned, you probably want to disable `Discovering` by default, because this is what this changes.

dbusmock/templates/bluez5.py
```diff
@@ -146,19 +195,22 @@ def AddDevice(self, adapter_device_name, device_address, alias):
...
-        'LegacyPairing': dbus.Boolean(False, variant_level=1),
-        'Paired': dbus.Boolean(False, variant_level=1),
-        'Trusted': dbus.Boolean(False, variant_level=1),
-        'RSSI': dbus.Int16(-79, variant_level=1),  # arbitrary
-        'Adapter': dbus.ObjectPath(adapter_path, variant_level=1),
-        'Address': dbus.String(device_address, variant_level=1),
-        'Alias': dbus.String(alias, variant_level=1),
-        'Name': dbus.String(alias, variant_level=1),
-    }
+    if path in microbit_data.keys():
```
hadess, line 198: I don't think we should have microbit specific code in here.

dbusmock/templates/bluez5.py
```diff
@@ -186,7 +239,7 @@ def AddDevice(self, adapter_device_name, device_address, alias):
 
 @dbus.service.method(BLUEZ_MOCK_IFACE,
                      in_signature='ssi', out_signature='')
-def PairDevice(_self, adapter_device_name, device_address, class_=5898764):
+def PairDevice(self, adapter_device_name, device_address, class_=5898764):
```
hadess, line 242: Why rename `_self` to `self`? If it's needed, it should probably happen in a separate commit.

dbusmock/templates/bluez5.py
```diff
@@ -393,3 +498,625 @@ def DisconnectDevice(_self, adapter_device_name, device_address):
...
+    manager.EmitSignal(OBJECT_MANAGER_IFACE, 'InterfacesAdded',
+                       'oa{sa{sv}}', [
+                           dbus.ObjectPath(path),
+                           {DEVICE_IFACE: properties},
+                       ])
+
+
+@dbus.service.method(BLUEZ_MOCK_IFACE,
+                     in_signature='', out_signature='')
+def DeviceDiscovery(self):
```
hadess, line 575: I don't think that we should be making beacons or devices appear by default when a discovery starts. It might however be a good idea to have a list of devices/beacons that would be discovered, which you could add to using a method on the Mock interface.

Each one of those beacon types you list here could be a separate beacon one could add, something like:
```
def AddDiscoveredDevice(self, beacon_type):
    if beacon_type == 'eddystone-url':
...
```

dbusmock/templates/bluez5.py
```diff
@@ -393,3 +498,625 @@ def DisconnectDevice(_self, adapter_device_name, device_address):
...
+                          'sa{sv}as', [
+                              GATT_CHRC_IFACE,
+                              {
+                                  'Value': dbus.Array(value, variant_level=1),
+                              },
+                              [],
+                          ])
+
+
+microbit_data = {
```
hadess, line 666: This really needs to be done some other way...

hadess (changes requested):

hadess:
@martinpitt is the one that can help with some of the specifics, hopefully

dbusmock/templates/bluez5.py
martinpitt: Agreed,  please drop this. See commit 801e10abb8

dbusmock/templates/bluez5.py
martinpitt: Right, it's a sign that it's an unused formal parameter. Please keep.

dbusmock/templates/bluez5.py
```diff
@@ -112,12 +130,43 @@ def AddAdapter(self, device_name, system_name):
         ('Register', 'ss', '', ''),
         ('Unregister', 's', '', ''),
     ])
+    adapter.AddMethods(LEDADVERTISING_MNGR_IFACE, [
+        ('RegisterAdvertisement', 'oa{sv}', '', ''),
+        ('UnregisterAdvertisement', 'o', '', '')
+    ])
```
martinpitt, line 136: Is this and the following method used by anything in `bluetoothctl`? Can you please add at least some cursory test to tests/test_bluez5.py to make sure that the API here is what the official tools expect?

dbusmock/templates/bluez5.py
```diff
@@ -393,3 +498,625 @@ def DisconnectDevice(_self, adapter_device_name, device_address):
...
+                   manf_id=76,
+                   manf_data=[2, 21, 106, 177, 124, 23, 244, 123, 77, 65, 128,
+                              54, 82, 106, 238, 210, 47, 115, 1, 22, 3, 104,
+                              191])
+
+
+@dbus.service.method(BLUEZ_MOCK_IFACE,
+                     in_signature='ss', out_signature='')
+def ConnectMicroBit(self, adapter_name, device_address):
+    print('In connect microbit')
```
martinpitt, line 603: No `print()` please, it may confuse or at least mess up the output of tests. Use `logging.debug()`.

dbusmock/templates/bluez5.py
martinpitt: Seconded, please keep out this giant blurb of magic numbers. This really belongs into the test suite that *uses* the dbusmock. Note that these are *templates*, and test suites are expect to configure/modify it as required for a particular test. This should provide a skeleton, not a set of highly specific behaviour.

martinpitt (changes requested):
Thanks @hadess for your great review! Big echo, and some additional comments. Thanks @ukBaz for improving the template!

Please also rebase to current master, @hadess made some changes there.

ukBaz:
Thanks @hadess and @martinpitt for the review. I guess I wasn't so clear on what I was looking for feedback on.

I agree that adding my specific data into the dbusmock library is wrong. What I was hoping to get feedback on was how I separate those two.
I want to move the BlueZ mock piece upstream and back into dbusmock. The second piece I want in my test suite. What I am struggling to see is how I separate those two things out. To add the data, I have to take a copy of the template into my project and then add data. At that point trying to keep my template in sync with the dbusmock template is difficult.

I'll close this PR and have a think again about how I separate the functions from the data.

martinpitt:
@ukBaz : What's wrong with putting the new API and generic structure into the template, and inject the specific data into the mock for particular tests? You don't need to copy the template file, you can add/change the mock at runtime through the `AddMethod()` etc. D-Bus calls; that's possible from pretty much any programming language. See the [README](https://github.com/martinpitt/python-dbusmock/blob/master/README.rst) for several examples. Setting up mocks dynamically is dbusmock's main mode -- the templates were originally meant to be a convenience add-on (I never imagined them to become that complex and functional :grin: )

PR #66, opened 2021-02-16
bluez: Implement Pair() method on the Device interface

hadess:
Tested successfully against gnome-bluetooth.


hadess:
Note that it doesn't implement using agents (yet?)

martinpitt (approved):
Merci beaucoup ! It'd be nice if that could be tested with bluetoothctl somehow, but as you run them in gnome-bluetooth, that's ok.

PR #68, opened 2021-03-15
template/accounts_service: Add accounts service mock implementation

3v1n0:
template/accounts_service: add accounts service mock implementation

it allows to create fake users to be added to the accountsservice,
implementing all the service methods.

all the user properties can be simply mocked calling the adduser mock method or
during initialization via parameters.

---

Added unit tests, however I've not added the required dependency (`gir1.2-accountsservice-1.0`) because there are two issues blocking such tests: (mainly [!61](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/65) and [!65](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/65))

### DIFF
--- a/dbusmock/templates/accounts_service.py
+++ b/dbusmock/templates/accounts_service.py
@@ -1,0 +1,388 @@
+'''Accounts Service D-Bus mock template'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Marco Trevisan'
+__email__ = 'marco.trevisan@canonical.com'
+__copyright__ = '(c) 2021 Canonical Ltd.'
+__license__ = 'LGPL 3+'
+
+import sys
+import time
+
+import dbus
+from dbusmock import MOCK_IFACE, mockobject
+
+BUS_NAME = 'org.freedesktop.Accounts'
+MAIN_OBJ = '/org/freedesktop/Accounts'
+MAIN_IFACE = 'org.freedesktop.Accounts'
+USER_IFACE = MAIN_IFACE + '.User'
+SYSTEM_BUS = True
+
+DEFAULT_USER_PASSWORD = 'Pa$$wo0rd'
+
+
+def get_user_path(uid):
+    return '/org/freedesktop/Accounts/User{}'.format(uid)
+
+
+def load(mock, parameters=None):
+    parameters = parameters if parameters else {}
+    mock.mock_users = {}
+    mock.cached_users = []
+    mock.automatic_login_users = set()
+    mock.users_auto_uids = 2000
+
+    mock.AddProperties(MAIN_IFACE, mock.GetAll(MAIN_IFACE))
+
+    for uid, name in parameters.get('users', {}).items():
+        mock.AddUser(uid, name)
+
+
+def emit_properties_changed(mock, interface=MAIN_IFACE, properties=None):
+    if properties is None:
+        properties = mock.GetAll(interface)
+    elif isinstance(properties, str):
+        properties = [properties]
+
+    if isinstance(properties, (list, set)):
+        properties = {p: mock.Get(interface, p) for p in properties}
+    elif not isinstance(properties, dict):
+        raise TypeError('Unsupported properties type')
+
+    mock.EmitSignal(dbus.PROPERTIES_IFACE, 'PropertiesChanged', 'sa{sv}as', (
+        interface, properties, []))
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='xssa{sv}a{sv}',
+                     out_signature='o')
+def AddUser(self, uid, username, password=DEFAULT_USER_PASSWORD,
+            overrides=None, password_policy_overrides=None):
+    '''Add user via uid and username and optionally overriding properties
+
+    Returns the new object path.
+   '''
+    path = get_user_path(uid)
+    default_props = {
+        'Uid': dbus.UInt64(uid),
+        'UserName': username,
+        'RealName': username[0].upper() + username[1:] + ' Fake',
+        'AccountType': dbus.Int32(1),
+        'AutomaticLogin': False,
+        'BackgroundFile': '',
+        'Email': '{}@python-dbusmock.org'.format(username),
+        'FormatsLocale': 'C',
+        'HomeDirectory': '/nonexisting/mock-home/{}'.format(username),
+        'IconFile': '',
+        'InputSources': dbus.Array([], signature='a{ss}'),
+        'Language': 'C',
+        'LocalAccount': True,
+        'Location': '',
+        'Locked': False,
+        'LoginFrequency': dbus.UInt64(0),
+        'LoginHistory': dbus.Array([], signature='(xxa{sv})'),
+        'LoginTime': dbus.Int64(0),
+        'PasswordHint': 'Remember it, come on!',
+        'PasswordMode': 0,
+        'Session': 'mock-session',
+        'SessionType': 'wayland',
+        'Shell': '/usr/bin/zsh',
+        'SystemAccount': False,
+        'XHasMessages': False,
+        'XKeyboardLayouts': dbus.Array([], signature='s'),
+        'XSession': 'mock-xsession',
+    }
+    default_props.update(overrides if overrides else {})
+    self.AddObject(path, USER_IFACE, default_props, [])
+
+    had_users = len(self.mock_users) != 0
+    had_multiple_users = len(self.mock_users) > 1
+    user = mockobject.objects[path]
+    user.password = password
+    user.properties = default_props
+    user.pwd_expiration_policy = {
+        'expiration_time': sys.maxsize,
+        'last_change_time': int(time.time()),
+        'min_days_between_changes': 0,
+        'max_days_between_changes': 0,
+        'days_to_warn': 0,
+        'days_after_expiration_until_lock': 0,
+    }
+    user.pwd_expiration_policy.update(
+        password_policy_overrides if password_policy_overrides else {})
+    self.mock_users[uid] = default_props
+
+    self.EmitSignal(MAIN_IFACE, 'UserAdded', 'o', [path])
+
+    if not had_users:
+        emit_properties_changed(self, MAIN_IFACE, 'HasNoUsers')
+    elif not had_multiple_users and len(self.mock_users) > 1:
+        emit_properties_changed(self, MAIN_IFACE, 'HasMultipleUsers')
+
+    return path
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='', out_signature='ao')
+def ListMockUsers(self):
+    """ List the mock users that have been created """
+    return [get_user_path(uid) for uid in self.mock_users.keys()]
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='s', out_signature='o')
+def AddAutoLoginUser(self, username):
+    """ Enable autologin for an user """
+    path = self.FindUserByName(username)
+    self.automatic_login_users.add(path)
+    user = mockobject.objects[path]
+    set_user_property(user, 'AutomaticLogin', True)
+    emit_properties_changed(self, MAIN_IFACE, 'AutomaticLoginUsers')
+    return path
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='s', out_signature='o')
+def RemoveAutoLoginUser(self, username):
+    """ Disables autologin for an user """
+    path = self.FindUserByName(username)
+    self.automatic_login_users.remove(path)
+    user = mockobject.objects[path]
+    set_user_property(user, 'AutomaticLogin', False)
+    emit_properties_changed(self, MAIN_IFACE, 'AutomaticLoginUsers')
+    return path
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='ssi', out_signature='o')
+def CreateUser(self, name, fullname, account_type):
+    """ Creates an user using the default API """
+    try:
+        self.FindUserByName(name)
+        found = True
+    except dbus.exceptions.DBusException:
+        found = False
+
+    if found:
+        raise dbus.exceptions.DBusException(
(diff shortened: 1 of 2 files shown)


martinpitt:
@3v1n0 : You still seem to iterate on this a lot. Moving to draft now, can you please let me know when you are done and want a review? Thanks!

3v1n0:
@martinpitt yeah, sorry I noticed few changes while doing proper testing on this, preparing https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/66

I guess it's fine now though

martinpitt:
@3v1n0 : Do you mind adding the gir-accountsservice dependency to tests/run-{debian,fedora} for the current releases, so that the tests actually run? Thanks!

martinpitt:
@3v1n0 : I.e. I thought the two "FIXMEs" would refer to the two outstanding PRs, and these checks are already commented out?

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+        'SystemAccount': False,
+        'XHasMessages': False,
+        'XKeyboardLayouts': dbus.Array([], signature='s'),
+        'XSession': 'mock-xsession',
+    }
+    default_props.update(overrides if overrides else {})
+    self.AddObject(path, USER_IFACE, default_props, [
+        ('SetRealName', 's', '', SetRealName),
+        ('SetEmail', 's', '', SetEmail),
+        ('SetLanguage', 's', '', SetLanguage),
```
martinpitt, line 97: These trivial setters could just be written inline here. There's a lot of them, they all look the same, and with full-blown functions they create a lot of redundancy and overhead (like duplicating the signature).

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+        ('SetPassword', 'ss', '', SetPassword),
+        ('SetAutomaticLogin', 'b', '', SetAutomaticLogin),
+        ('GetPasswordExpirationPolicy', '',
+         'xxxxxx', GetPasswordExpirationPolicy),
+    ])
+
+    had_users = len(self.mock_users) != 0
+    had_multiple_users = len(self.mock_users) > 1
+    user = mockobject.objects[path]
+    user.password = 'Pa$$wo0rd'
```
martinpitt, line 118: Please define this as a `DEFAULT_PASSWORD` constant at the top of the file, so that it's easier to spot.

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+    user.properties = default_props
+    user.pwd_expiration_policy = {
+        'expiration_time': 1000,
+        'last_change_time': 1615832413,
```
martinpitt, line 122: Magic number. Please define as a constant and add a comment what it is in ISO time format. Why does it have to be a fixed hardcoded time? Would it make more sense to be relative to the current time?

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+    default_props = {
+        'Uid': dbus.UInt64(uid),
+        'UserName': username,
+        'RealName': username[0].upper() + username[1:] + ' Fake',
+        'AccountType': dbus.Int32(1),
+        'AutomaticLogin': False,
+        'BackgroundFile': '',
+        'Email': '{}@python-dbusmock.org'.format(username),
+        'FormatsLocale': 'C',
+        'HomeDirectory': '/tmp/dbusmock-home/{}'.format(username),
```
martinpitt, line 73: The path could also do with a DEFAULT_ constant. Why this odd path? Is it reasonable that test suites would actually create/write into this? Then it shouldn't be a predictable dir in /tmp/, as that's always prone to symlink attacks, existing files, etc. Make it a proper `mkdtemp()`, or something like `/mock/home/` if this is never actually written.

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='', out_signature='ao')
+def ListMockUsers(self):
+    """ List the mock users that have been created """
+    return [get_user_path(uid) for uid in self.mock_users.keys()]
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='s', out_signature='o')
+def AddAutoLoginUser(self, username):
```
martinpitt, line 147: It surprises me that accountsservice has no official API to set this?

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+    """ Enable autologin for an user """
+    path = self.FindUserByName(username)
+    self.automatic_login_users.append(path)
```
martinpitt, line 150: Indeed I now saw that accountsservice has an `ao` property `AutomaticLoginUsers`, so the mock is faithful to that -- still it surprises me that this makes sense.

However, this should certainly have some error handlng if that user is already in the list?

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+    mockobject.objects[path].SetAutomaticLogin(True)
+    emit_properties_changed(self, MAIN_IFACE, 'AutomaticLoginUsers')
+    return path
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='s', out_signature='o')
+def RemoveAutoLoginUser(self, username):
+    """ Disables autologin for an user """
+    path = self.FindUserByName(username)
+    self.automatic_login_users.remove(path)
```
martinpitt, line 160: Would crash with ValueError if the user does not have autologin. This is most certainly not what the real accountsservice does. Please add a unit test for the expected behaviour.

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+    emit_properties_changed(self, MAIN_IFACE, 'AutomaticLoginUsers')
+    return path
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='ssi', out_signature='o')
+def CreateUser(self, name, fullname, account_type):
+    """ Creates an user using the default API """
+    try:
+        self.FindUserByName(name)
+        found = True
```
martinpitt, line 171: Please raise the exception directly here and eliminate `found`.

tests/test_accounts_service.py
```diff
@@ -0,0 +1,239 @@
...
+            changed = u is user
+        conn_id = user.connect('changed', on_changed)
+        while not changed:
+            self.ctx.iteration(True)
+        user.disconnect(conn_id)
+
+    def test_empty(self):
+        self.assertTrue(self.p_manager.props.is_loaded)
+        self.assertFalse(self.p_manager.list_users())
+        self.assertFalse(self.p_manager.props.has_multiple_users)
```
martinpitt, line 83: check has_no_users as well here?

tests/test_accounts_service.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def test_create_users(self):
+        self.p_manager.create_user(
+            'pizza', 'I Love Pizza',
+            AccountsService.UserAccountType.ADMINISTRATOR)
+        self.assertFalse(self.p_manager.props.has_multiple_users)
```
martinpitt, line 89: likewise, has_no_users should be false now.

tests/test_accounts_service.py
```diff
@@ -0,0 +1,239 @@
...
+
+        user.set_account_type(
+            AccountsService.UserAccountType.STANDARD)
+        self.wait_changed(user)
+        self.assertEqual(user.get_account_type(),
+                         AccountsService.UserAccountType.STANDARD)
+
+        user.set_automatic_login(True)
+        self.wait_changed(user)
+        self.assertTrue(user.get_automatic_login())
```
martinpitt, line 164: does the client library also expose `AutomaticLoginUsers`? If so, that should be checked, too.

tests/test_accounts_service.py
```diff
@@ -0,0 +1,239 @@
...
+
+        user.set_automatic_login(False)
+        self.wait_changed(user)
+        self.assertFalse(user.get_automatic_login())
+
+        user.set_email('Test Email')
```
martinpitt, line 170: let's please use a validly formed email, both to make this more readable and also prevent failure if the daemon or the library ever starts to check this.

martinpitt (changes requested):
Thanks! Lots of little nitpicks, but nothing earthshattering.

3v1n0:
> @3v1n0 : Do you mind adding the gir-accountsservice dependency to tests/run-{debian,fedora} for the current releases, so that the tests actually run? Thanks!

I wrote the rationale in the PR description, as we can't rely on the released versions yet :(

> Added unit tests, however I've not added the required dependency (gir1.2-accountsservice-1.0) because there are two issues blocking such tests: (mainly !61 and !65)

dbusmock/templates/accounts_service.py
```diff
@@ -0,0 +1,392 @@
...
+def emit_properties_changed(mock, interface=MAIN_IFACE, properties=None):
+    if properties is None:
+        properties = mock.GetAll(interface)
+    elif isinstance(properties, str):
+        properties = [properties]
+
+    if isinstance(properties, (list, set)):
+        properties = {p: mock.Get(interface, p) for p in properties}
+    elif isinstance(properties, dict):
+        pass
```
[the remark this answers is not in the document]
3v1n0: Yeah, was just to make it clearer what we expect, I can instead just raise indeed.

dbusmock/templates/accounts_service.py
3v1n0: Ok, I tend to prefer to use real methods as they have some better static checks imho, so we can avoid duplications via https://github.com/martinpitt/python-dbusmock/pull/72 (and jonas utility method to just add a method via its handle), avoiding all the duplication I guess, no?

dbusmock/templates/accounts_service.py
3v1n0: All this mock started as a simple tool for testing fprintd locally so didn't bother much in having real scenarios, ideally I just wanted to be to a location that never existed, not sure we should really create a real one at this level though, maybe we can just unset it and leave it to be handled by those who are using the template by overriding if needed?

tests/test_accounts_service.py
3v1n0: That's not exported via Gir, so in fact I was thinking to test this via the actual dbus api instead. I forgot to do it though.

tests/test_accounts_service.py
3v1n0: Nope, that's not.

I should probably just test it via dbus though. In any case while I've been quite through with testing the internals of the lib here, I actually added more tests into accountsservice repo itself (where they belong more).

I tried here to test what we exposed via the mock though, so in this case I think the only way is going through dbus again

martinpitt: Thanks; no worries then -- tests against dbus are very shallow, as they don't test any "official" specification. They merely tell you that your code doesn't crash. So that's fine then.

tests/test_accounts_service.py
martinpitt: No, that's fine --  just check the official lib API.

dbusmock/templates/accounts_service.py
martinpitt: Ack; if it's never written to, please clarify that with /mock/home or /nonexisting/home or something such.

dbusmock/templates/accounts_service.py
martinpitt: I'll review #72, thanks. I still don't get why you want/need to *both* enumerate them here *and* define them as proper functions. Please pick just one. I don't understand what "better static checks" a trivial function like
```
@dbus.service.method(USER_IFACE, in_signature='s')
def SetEmail(self, email):
    set_user_property(self, 'Email', email)
```
does -- that does exactly the same (i.e. hardly any) static checks than

    ('SetEmail', 's', '', 'self.set_user_property("Email", args[0])')

and avoids the boilerplate. But anyway, if you prefer the full-blown functions I don't veto that, but then please remove them from the AddObject() call.

3v1n0: > and avoids the boilerplate. But anyway, if you prefer the full-blown functions I don't veto that, but then please remove them from the AddObject() call.

Ohhh... So, I thought that this was required in order to make them work, for some reason I think they didn't work when I initially did this, thus the reason of listing them (and eventually of #72).

dbusmock/templates/accounts_service.py
3v1n0: Again, was just adding random stuff around, will make them a bit more real.

dbusmock/templates/accounts_service.py
3v1n0: Eh, I try to do that previously, but the problem here is that I want to proceed if the exception is raised, so if I raise a new one here I'll be caught by the `except dbus.exceptions.DBusException` later on, so... I'd never be able to handle this.

I was looking to some clearner way to do this but didn't get anything better than this

martinpitt: Heh, right, nevermind then.

martinpitt:
Thanks, this now looks much cleaner! There are still two outstanding review comments about proper error handling (they are hidden by default, you need to expand them..) But this looks almost ready then. So the unit tests succeed against current accountsservice git master? Or are there still other outstanding bugs/PRs?

3v1n0:
> (they are hidden by default, you need to expand them..)

Oh, sorry, I missed they were hidden, but it looks like I've addressed most of them anyways by myself, I'll double-check though.

> So the unit tests succeed against current accountsservice git master? Or are there still other outstanding bugs/PRs?

master is good, not sure if you want to install from it in CI instead... Let me know.

dbusmock/templates/accounts_service.py
3v1n0: I'm using a set now, so... We can error indeed if called via the mock though.

dbusmock/templates/accounts_service.py
3v1n0: This is the mock iface so I'd actually prefer it to throw, so the reason I kept it as is.

martinpitt (approved):
Thanks for the follow-ups! So this seems good enough to land, and then I'll do an upstream release so that you and @jadahl can use the new toys :grin:

PR #69, opened 2021-03-16
templates/polkitd: Also mock 'RegisterAuthenticationAgent()'

jadahl:
Part of an attempt to run GNOME Shell with python-dbusmock

### DIFF
--- a/dbusmock/templates/polkitd.py
+++ b/dbusmock/templates/polkitd.py
@@ -26,18 +26,6 @@ SYSTEM_BUS = True
 
 
 def load(mock, _parameters):
-    mock.AddMethod(MAIN_IFACE,
-                   'CheckAuthorization',
-                   '(sa{sv})sa{ss}us',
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thank you! Did that work for you, and is it complete enough, or are you still working on this?

jadahl:
> Thank you! Did that work for you, and is it complete enough, or are you still working on this?

It did (at least I didn't see any complaint about anything more in the logs), but I think @3v1n0 had something he thought could be added here too, so maybe wait a bit.

3v1n0:
Yeah so since we were here, we've some changes in fprintd that would be nice to upstream, so here they are: https://github.com/jadahl/python-dbusmock/pull/1/

jadahl:
@martinpitt this now has @3v1n0 changes too, so should be ready.

martinpitt (approved):
Very cool, thanks!

PR #70, opened 2021-03-16
templates: Add evolution-data-server template

jadahl:
The template is seemingly empty, but the important part is that it is an object
manager.

-- 

Another one added to run gnome-shell with dbus-mock.

### DIFF
--- a/dbusmock/templates/evolution-data-server.py
+++ b/dbusmock/templates/evolution-data-server.py
@@ -1,0 +1,22 @@
+'''evolution-data-server mock template
+'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Jonas Ådahl'
+__copyright__ = '(c) Red Hat'
+
+
+BUS_NAME = 'org.gnome.evolution.dataserver.Sources5'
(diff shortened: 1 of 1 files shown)


martinpitt:
Honestly I don't like that much. It doesn't seem like a mock that is widely used (unlike polkit and such), and it doesn't have any interesting functionality. Specific things like these would IMHO better live in e.g. evolution or gnome-shell's integration tests directly, either as inline code or as local template. Then they can also be modified for each test case separately.

jadahl:
Well it could potentially be used by e.g. `gnome-calendar` or anything other that wants to hook into calendar events, but I see your point. I added it here because the template could get "ObjectManager" functionality, and I couldn't see how I could get that without using templates.

I really would like to avoid having to manually write a mock evolution data server, as this fairly simple "implementation" was enough to get by (just means there are no calendar, which is fine). Can I use the python-dbusmock object manager without a template upstream?

3v1n0:
> I really would like to avoid having to manually write a mock evolution data server, as this fairly simple "implementation" was enough to get by (just means there are no calendar, which is fine). Can I use the python-dbusmock object manager without a template upstream?

You can also just use the template locally, you've just do do something [as we do in other projects](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/blob/master/tests/test-libaccountsservice.py#L41-44).

jadahl:
> You can also just use the template locally, you've just do do something [as we do in other projects](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/blob/master/tests/test-libaccountsservice.py#L41-44).

Ah, that's perfect, thanks.

PR #71, opened 2021-03-16
templates: Add systemd-user and systemd-system templates

jadahl:
The templates shares a common implementation, in systemd-base.py, that
implements mocked versions of StartUnit, StartTransientUnit and
StopUnit. It imitates systemd enough to make gnome-shell happy, thinking
it managed to launch the units it asked to have launched.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -88,6 +88,36 @@ def _format_args(args):
     return s
 
 
+def _wrap_in_dbus_variant(value):
+    dbus_types = [
+        dbus.types.Array,
+        dbus.types.ByteArray,
+        dbus.types.Int16,
+        dbus.types.ObjectPath,
+        dbus.types.Struct,
+        dbus.types.UInt64,
+        dbus.types.Boolean,
+        dbus.types.Dictionary,
+        dbus.types.Int32,
+        dbus.types.Signature,
+        dbus.types.UInt16,
+        dbus.types.UnixFd,
+        dbus.types.Byte,
+        dbus.types.Double,
+        dbus.types.Int64,
+        dbus.types.String,
+        dbus.types.UInt32,
+    ]
+    if isinstance(value, dbus.String):
+        return dbus.String(str(value), variant_level=1)
+    if type(value) in dbus_types:
+        return type(value)(value.conjugate(), variant_level=1)
+    if isinstance(value, str):
+        return dbus.String(value, variant_level=1)
+    raise dbus.exceptions.DBusException(
+        'could not wrap type {}'.format(type(value)))
+
+
 class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-attributes
     '''Mock D-Bus object
 
@@ -400,6 +430,42 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         for method in methods:
             self.AddMethod(interface, *method)
 
+    def _set_property(self, interface, name, value):
+        # copy.copy removes one level of variant-ness, which means that the
+        # types get exported in introspection data correctly, but we can't do
+        # this for container types.
+        if not isinstance(value, (dbus.Dictionary, dbus.Array)):
+            value = copy.copy(value)
+
+        self.props.setdefault(interface, {})[name] = value
+
+    @dbus.service.method(MOCK_IFACE,
+                         in_signature='sa{sv}',
+                         out_signature='')
+    def UpdateProperties(self, interface: str, properties: PropsType) -> None:
+        '''Update properties on this object and send a PropertiesChanged signal
+
+        interface: D-Bus interface to update this to. For convenience you can
+                   specify '' here to add the property to the object's main
+                   interface (as specified on construction).
+        properties: A property_name (string) → value map
+        '''
+        changed_props = {}
+
+        for name, value in properties.items():
+            if not interface:
+                interface = self.interface
+            if name not in self.props.get(interface, {}):
+                raise dbus.exceptions.DBusException(
+                    'property %s not found' % name,
+                    name=interface + '.NoSuchProperty')
+
+            self._set_property(interface, name, value)
+            changed_props[name] = _wrap_in_dbus_variant(value)
+
+        self.EmitSignal(dbus.PROPERTIES_IFACE, 'PropertiesChanged', 'sa{sv}as', [
+            interface, changed_props, []])
+
     @dbus.service.method(MOCK_IFACE,
                          in_signature='ssv',
                          out_signature='')
@@ -419,13 +485,7 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
                 'property %s already exists' % name,
                 name=self.interface + '.PropertyExists')
 
-        # copy.copy removes one level of variant-ness, which means that the
-        # types get exported in introspection data correctly, but we can't do
-        # this for container types.
-        if not isinstance(value, (dbus.Dictionary, dbus.Array)):
-            value = copy.copy(value)
-
-        self.props.setdefault(interface, {})[name] = value
+        self._set_property(interface, name, value)
 
     @dbus.service.method(MOCK_IFACE,
                          in_signature='sa{sv}',
--- a/dbusmock/templates/systemd.py
+++ b/dbusmock/templates/systemd.py
@@ -1,0 +1,106 @@
+'''systemd mock template
+'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Jonas Ådahl'
+__copyright__ = '(c) 2021 Red Hat'
+
+from gi.repository import GLib
+import dbus
+
+from dbusmock import MOCK_IFACE, mockobject
+
+BUS_PREFIX = 'org.freedesktop.systemd1'
+PATH_PREFIX = '/org/freedesktop/systemd1'
+
+
+BUS_NAME = BUS_PREFIX
+MAIN_OBJ = PATH_PREFIX
+MAIN_IFACE = BUS_PREFIX + '.Manager'
+UNIT_IFACE = BUS_PREFIX + '.Unit'
+SYSTEM_BUS = True
+
+
+def load(mock, _parameters):
+    mock.next_job_id = 1
+    mock.units = {}
+
+    mock.AddProperties(MAIN_IFACE, {'Version': 'v246'})
+
+
+def escape_unit_name(name):
+    for s in ['.', '-']:
+        name = name.replace(s, '_')
+    return name
+
+
+def emit_job_new_remove(mock, job_id, job_path, name):
+    mock.EmitSignal(MAIN_IFACE, 'JobNew', 'uos', [job_id, job_path, name])
+    mock.EmitSignal(MAIN_IFACE, 'JobRemoved', 'uoss',
+                    [job_id, job_path, name, 'done'])
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='ss', out_signature='o')
+def StartUnit(self, name, _mode):
+    job_id = self.next_job_id
+    self.next_job_id += 1
+
+    job_path = PATH_PREFIX + '/Job/{}'.format(job_id)
+    GLib.idle_add(lambda: emit_job_new_remove(self, job_id, job_path, name))
+
+    unit_path = self.units[str(name)]
+    unit = mockobject.objects[unit_path]
+    unit.UpdateProperties(UNIT_IFACE, {'ActiveState': 'active'})
+
+    return job_path
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='ssa(sv)a(sa(sv))', out_signature='o')
+def StartTransientUnit(self, name, _mode, _properties, _aux):
+    job_id = self.next_job_id
+    self.next_job_id += 1
+
+    job_path = PATH_PREFIX + '/Job/%d' % (job_id)
+    GLib.idle_add(lambda: emit_job_new_remove(self, job_id, job_path, name))
+
+    return job_path
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='ss', out_signature='o')
+def StopUnit(self, name, _mode):
+    job_id = self.next_job_id
+    self.next_job_id += 1
+
+    job_path = PATH_PREFIX + '/Job/%d' % (job_id)
+    GLib.idle_add(lambda: emit_job_new_remove(self, job_id, job_path, name))
+
+    unit_path = self.units[str(name)]
+    unit = mockobject.objects[unit_path]
+    unit.UpdateProperties(UNIT_IFACE, {'ActiveState': 'inactive'})
+    return job_path
+
+
+@dbus.service.method(MAIN_IFACE, in_signature='s', out_signature='o')
+def GetUnit(self, name):
+    unit_path = self.units[str(name)]
+    return unit_path
+
+
+@dbus.service.method(MOCK_IFACE, in_signature='s')
+def AddMockUnit(self, name):
+    unit_path = PATH_PREFIX + '/unit/%s' % (escape_unit_name(name))
+    self.units[str(name)] = unit_path
+    self.AddObject(unit_path,
+                   UNIT_IFACE,
+                   {
+                       'Id': name,
+                       'Names': [name],
+                       'LoadState': 'loaded',
+                       'ActiveState': 'inactive',
+                   },
+                   [])
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -338,6 +338,43 @@ assert args[2] == 5
         self.assertEqual(self.dbus_props.Get('org.freedesktop.Test.Other', 'color'),
                          'yellow')
 
+        changed_props = []
+        ml = GLib.MainLoop()
+
+        def catch(*args, **kwargs):
+            if kwargs['interface'] != 'org.freedesktop.DBus.Properties':
+                return
+
+            self.assertEqual(kwargs['interface'], 'org.freedesktop.DBus.Properties')
+            self.assertEqual(kwargs['member'], 'PropertiesChanged')
+
+            [iface, changed, _invalidated] = args
+            self.assertEqual(iface, 'org.freedesktop.Test.Main')
+
+            changed_props.append(changed)
+            ml.quit()
+
+        match = self.dbus_con.add_signal_receiver(catch,
+                                                  interface_keyword='interface',
+                                                  path_keyword='path',
+                                                  member_keyword='member')
+
+        # change property using mock helper
+        self.dbus_mock.UpdateProperties('org.freedesktop.Test.Main', {
+            'version': 5,
+            'connected': False,
+        })
+
+        GLib.timeout_add(3000, ml.quit)
+        ml.run()
+
+        match.remove()
+
+        self.assertEqual(self.dbus_props.GetAll('org.freedesktop.Test.Main'),
+                         {'version': 5, 'connected': False})
+        self.assertEqual(changed_props,
+                         [{'version': 5, 'connected': False}])
+
     def test_introspection_methods(self):
         '''dynamically added methods appear in introspection'''
 
--- a/tests/test_systemd.py
+++ b/tests/test_systemd.py
@@ -1,0 +1,112 @@
+#!/usr/bin/python3
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Jonas Ådahl'
+__copyright__ = '(c) 2021 Red Hat'
+
+import subprocess
+import sys
+import unittest
+import dbus
+import dbus.mainloop.glib
+
+from gi.repository import GLib
+
+import dbusmock
+
+dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+
+
+class TestSystemd(dbusmock.DBusTestCase):
+    '''Test mocking systemd'''
+
+    @classmethod
+    def setUpClass(cls):
+        cls.start_session_bus()
+        cls.start_system_bus()
+        cls.session_bus = cls.get_dbus(False)
+        cls.system_bus = cls.get_dbus(True)
+
+    def setUp(self):
+        self.p_mock = None
+
+    def tearDown(self):
+        if self.p_mock:
+            self.p_mock.stdout.close()
+            self.p_mock.terminate()
+            self.p_mock.wait()
+
+    def _assert_unit_property(self, unit_obj, name, expect):
+        value = unit_obj.Get('org.freedesktop.systemd1.Unit', name)
+        self.assertEqual(str(value), expect)
+
+    def _test_base(self, bus, system_bus=True):
+        dummy_service = 'dummy-dbusmock.service'
+
+        (self.p_mock, obj_systemd) = self.spawn_server_template('systemd', {},
+                                                                subprocess.PIPE,
+                                                                system_bus=system_bus)
+
+        systemd_mock = dbus.Interface(obj_systemd, dbusmock.MOCK_IFACE)
+        systemd_mock.AddMockUnit(dummy_service)
+
+        main_loop = GLib.MainLoop()
+
+        removed_jobs = []
+
+        def catch_job_removed(*args, **kwargs):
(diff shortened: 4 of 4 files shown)


dbusmock/templates/systemd-sytsem.py
```diff
@@ -0,0 +1,29 @@
+'''systemd mock template
```
martinpitt, line 1: file name typo: "sytstem.py"

martinpitt (changes requested):
Can you please add a minimal integration test? I suppose the API isn't complete enough to convince `systemctl`, but using that would really be the best test -- the official client-side tools ought to work (within reason and possibility) against the mock. That may require a few extra enumerated properties, but that may not be too hard?

jadahl:
> That may require a few extra enumerated properties, but that may not be too hard?

You mean `systemctl start foo.service` should "work"? I could try, but I suspect `status` and `restart` etc will not; this was the amount of functionality needed to make gnome-shell believe it managed to start the `gnome-shell-x11-services.target`, which more or less was `StartUnit` and a `JobRemoved`, as well as `gnome-desktop`'s usage of `StartTransientUnit` to do.. something.

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def load_base(mock, _parameters):
+    mock.next_job_id = 1
+    mock.units = {}
+
+    mock.AddMethod(MAIN_IFACE,
+                   'StartUnit',
+                   'ss', 'o',
+                   'ret = self.StartUnit(self, args[0], args[1])')
```
3v1n0, line 27: No need to do this, just use the `@dbus.service.method(MAIN_IFACE, in_signature=ss', out_signature='o')` annotation on actual method, and you don't need this at all (or at least no implementation of it).

3v1n0: And on other methods as well...

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,91 @@
...
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Jonas Ådahl'
+__copyright__ = '(c) 2021 Red Hat'
+
+from gi.repository import GLib
+
+BUS_NAME = 'org.freedesktop.systemd1'
+MAIN_OBJ = '/org/freedesktop/systemd1'
```
3v1n0, line 16: Maybe you could use these as basic prefix for avoiding repetitions below?

dbusmock/templates/systemd-base.py
jadahl: That didn't work when the method was implemented in `systemd-base.py`.

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,91 @@
...
+    self.EmitSignal('org.freedesktop.systemd1.Manager', 'JobRemoved', 'uoss',
+                    [job_id, job_path, name, 'done'])
+
+
+def StartUnit(self, name, _mode):
+    job_id = self.next_job_id
+    self.next_job_id += 1
+
+    job_path = '/org/freedesktop/systemd1/Job/%d' % (job_id)
+    GLib.idle_add(lambda: emit_job_new_remove(self, job_id, job_path, name))
```
3v1n0, line 59: If you want to do this on request, you could also do the async calls way that as done in polkitd

3v1n0 (commented):
Looks good to me, a part would be better to have those methods.

I'm also wondering if it's possible to instead just have a `systemd` template and another `systemd_user` one that just implements the other without having the base (you replied to me in private when we were discussing about this, but not sure if Martin has an opinion on this)

dbusmock/templates/systemd-base.py
3v1n0: Oh, I see (even with no implementation at all, right?)... So, maybe you can still use argument expansion though?

jadahl: Argument expansion?

3v1n0:
```suggestion
                   'ret = self.StartUnit(self, *args)')
```
So you don't have to do it manually for all the calls.

jadahl: Using `*args` now.

> even with no implementation at all, right?)...

Initially I had the decorated methods working without having the implementation in a separate file, but once I moved it to `systemd-base.py` with `systemd-user.py` being the small layer, nothing worked anymore. Tried to find different ways to "merge" the -base.py and the -user.py some how, but gave up and did it like this.

dbusmock/templates/systemd-base.py
jadahl: Added prefixes for the name and path, and changed to use that in the rest of the file.

dbusmock/templates/systemd-base.py
jadahl: I left these as is, what matters is that they are emitted after returning, and this felt like the simplest way to achieve that.

jadahl:
So I ended up not using `systemctl` to test, as it continuously didn't use the D-Bus API at all, and just read service files from disk, tried to launch things itself, and so on, so I removed my then `dummys-dbusmock.service` file I had and just went and tested the D-Bus methods themself.

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,126 @@
...
+                    [job_id, job_path, name])
+    self.EmitSignal('org.freedesktop.systemd1.Manager', 'JobRemoved', 'uoss',
+                    [job_id, job_path, name, 'done'])
+
+
+def StartUnit(self, name, _mode):
+    job_id = self.next_job_id
+    self.next_job_id += 1
+
+    job_path = PATH_PREFIX + '/Job/%d' % (job_id)
```
3v1n0, line 77: Just a nit, but these days I guess it's better to use `.format()` instead

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,128 @@
...
+    GLib.idle_add(lambda: emit_job_new_remove(self, job_id, job_path, name))
+
+    unit_path = self.units[str(name)]
```
3v1n0, line 82: Maybe adding a fallback using `self.units.get(name, '/path/foobar/{}'.format(name))` so that it can be used as before without adding a mock unit (for conveience, without throwing a python error)?

Or instead on KeyError raise a proper dbus error as systemd would do?

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,128 @@
...
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Jonas Ådahl'
+__copyright__ = '(c) 2021 Red Hat'
+
+from gi.repository import GLib
+
+from dbusmock import MOCK_IFACE
+from dbusmock import mockobject
```
3v1n0, line 16: Guess you can merge those two lines

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,128 @@
...
+
+BUS_PREFIX = 'org.freedesktop.systemd1'
+PATH_PREFIX = '/org/freedesktop/systemd1'
+
+
+BUS_NAME = BUS_PREFIX
+MAIN_OBJ = PATH_PREFIX
+MAIN_IFACE = BUS_PREFIX + '.Manager'
+
+
```
3v1n0, line 26: Guess you can get rid of those too

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,128 @@
...
+                   'ret = self.GetUnit(self, *args)')
+    mock.AddProperties(MAIN_IFACE, {'Version': 'v246'})
+
+    mock.StartUnit = StartUnit
+    mock.StartTransientUnit = StartTransientUnit
+    mock.StopUnit = StopUnit
+    mock.GetUnit = GetUnit
+
+    mock.AddMethod(MOCK_IFACE, 'AddMockUnit', 's', '', 'self.AddMockUnit(self, *args)')
+
```
3v1n0, line 58: Sooo... This can be reduced by just using `AddMethods` and annotations (that will help us also in handling the code better) via https://paste.ubuntu.com/p/gJS9td7HSR/

You add an even better implementation without having to repeat the signature though :)

dbusmock/templates/systemd-base.py
jadahl: I was assuming key errors would make tests fail, so didn't bother with any helpful code here.

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,128 @@
...
+    for s in ['.', '-']:
+        name = name.replace(s, '_')
+    return name
+
+
+def emit_job_new_remove(self, job_id, job_path, name):
+    self.EmitSignal('org.freedesktop.systemd1.Manager', 'JobNew', 'uos',
+                    [job_id, job_path, name])
+    self.EmitSignal('org.freedesktop.systemd1.Manager', 'JobRemoved', 'uoss',
+                    [job_id, job_path, name, 'done'])
```
3v1n0, line 72:
```suggestion
    self.EmitSignal(MAIN_IFACE, 'JobNew', 'uos', [job_id, job_path, name])
    self.EmitSignal(MAIN_IFACE, 'JobRemoved', 'uoss',
                    [job_id, job_path, name, 'done'])
```

jadahl: FIxed

dbusmock/templates/systemd-base.py
jadahl: Did that.

dbusmock/templates/systemd-base.py
jadahl: We'll see if the linter complains now that I removed them.

dbusmock/templates/systemd-base.py
jadahl: Well, we'll see what @martinpitt thinks about the new mockobject method :P

dbusmock/templates/systemd-base.py
jadahl: Using format now.

dbusmock/templates/systemd-base.py
3v1n0: Well systemd will just emit

`org.freedesktop.DBus.Error.InvalidArgs: Unit name foo is not valid.`, so not hard to mimic it I guess.

dbusmock/templates/systemd-base.py
jadahl: Using prefixes now.

dbusmock/templates/systemd-base.py
jadahl: Did that.

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,111 @@
...
+    mock.AddServiceMethod(MOCK_IFACE, 'AddMockUnit', AddMockUnit)
+
+
+def escape_unit_name(name):
+    for s in ['.', '-']:
+        name = name.replace(s, '_')
+    return name
+
+
+def emit_job_new_remove(self, job_id, job_path, name):
```
3v1n0, line 47: Given this is not an object "class" function, maybe better not to use `self` here but just `mock` maybe?

jadahl: Hrm, none of the functions are in a class and use `self`..

dbusmock/mockobject.py
```diff
@@ -326,6 +356,21 @@ def Reset(self) -> None:
...
+        interface: D-Bus interface to add this to. For convenience you can
+                   specify '' here to add the method to the object's main
+                   interface (as specified on construction).
+        name: Name of the method
+        method: A method decorated using @dbus.service.method
+        '''
+        # pylint: disable=protected-access
+        self.AddMethod(interface, name,
+                       method._dbus_in_signature,
+                       method._dbus_out_signature,
```
3v1n0, line 371: Maybe we can protect here checking whether we're handling a function and if hasattr such values? Raising otherwise?

If doing this though, we could just maybe avoid adding another function and just add this as a special case of `AddMethod`.... Actually this will also solve another problem we have with decoration arguments so, could you instead pick from my branch?

https://github.com/3v1n0/python-dbusmock/commits/systemd-unit

dbusmock/templates/systemd-base.py
3v1n0: So... I've given this again a try, and I'd still prefer to just use pure python-dbus stuff here, and this can be done with just https://github.com/3v1n0/python-dbusmock/commit/dac4db16708a4dc8d52b9912bbc327c9840ed8ea which IMHO is still nicer (and in theory will allow to make it do a delayed if someone will need it in future).

dbusmock/mockobject.py
jadahl: I'll wait for @martinpitt to know whether he wants the regular AddMethod to handle this, and if so, I think the AddServiceMethod is a bit unnecessary.

3v1n0: In any case we can discuss that prior change in another PR, but I guess it's quite needed, as as explained many decorator parameters won't work otherwise.

dbusmock/templates/systemd-base.py
3v1n0: Not technically, but practically we consider them as part of the "mock dbus class", but not a big deal, I just was finding it clearer.

dbusmock/templates/systemd-base.py
3v1n0: I guess this can be marked as resolved, no idea how I can do it :stuck_out_tongue_closed_eyes:

dbusmock/mockobject.py
martinpitt: Honestly this feels a bit weird and bolted-on. MockObject is a D-Bus object itself, its API is via D-Bus. You can't call AddServiceMethod() through that. There is two ways to add a mocked method:

 1. through `AddMethod()` with the code as string; this works for any programming language (even at the shell), and is meant for dynamically shaping mocks in a project's unit tests.
 2. declaring a proper D-Bus function in the template [like here](https://github.com/martinpitt/python-dbusmock/blob/master/dbusmock/templates/logind.py#L82); this is for writing non-trivial methods in templates.

Why do you need this "hybrid" and indirect approach?

dbusmock/templates/systemd-base.py
```diff
@@ -0,0 +1,124 @@
...
+                   'StopUnit',
+                   'ss', 'o',
+                   'ret = self.StopUnit(self, *args)')
+    mock.AddMethod(MAIN_IFACE,
+                   'GetUnit',
+                   's', 'o',
+                   'ret = self.GetUnit(self, *args)')
+    mock.AddProperties(MAIN_IFACE, {'Version': 'v246'})
+
+    mock.StartUnit = StartUnit
```
martinpitt: Why this indirection? If you annotate `def StartUnit()` with ``@dbus.service.method`, it should be much simpler?

dbusmock/templates/systemd-system.py
```diff
@@ -0,0 +1,29 @@
...
+__copyright__ = '(c) 2021 Red Hat'
+
+from dbusmock import mockobject
+
+systemd_base = mockobject.load_module('systemd-base')
+
+BUS_NAME = systemd_base.BUS_NAME
+MAIN_OBJ = systemd_base.MAIN_OBJ
+MAIN_IFACE = systemd_base.MAIN_IFACE
+SYSTEM_BUS = True
```
martinpitt, line 25: That's the only place where the two differ, right? I suppose it would be nice if for these cases `spawn_server_template()` could be parameterized to launch the template on the system or session bus, e. g. by looking at `parameters.get('SYSTEM_BUS', module.SYSTEM_BUS)`, or by adding a new `system_bus=None` kwarg. Besides systemd, I can also think of libvirtd as a service which can legitimately appear on both buses.

If you think that will help you, I can get this in relatively quickly -- or of course feel free to include it in your branch.

dbusmock/templates/systemd-base.py
```diff
@@ -28,32 +29,13 @@ def load_base(mock, _parameters):
...
-                   'ret = self.StartTransientUnit(self, *args)')
-    mock.AddMethod(MAIN_IFACE,
-                   'StopUnit',
-                   'ss', 'o',
-                   'ret = self.StopUnit(self, *args)')
-    mock.AddMethod(MAIN_IFACE,
-                   'GetUnit',
-                   's', 'o',
-                   'ret = self.GetUnit(self, *args)')
+    mock.AddServiceMethod(MAIN_IFACE, 'StartUnit', StartUnit)
```
martinpitt, line 32: As this commit now actually adds the decorations, I wonder why you need these calls at all now? The decorated methods are supposed to "just work".

martinpitt (changes requested):
Thanks! Let me know what you think about the parameterizable system/session bus for loading templates to make this much simpler.

dbusmock/templates/systemd-base.py
jadahl: The indirection was added due to using the decoration did not work when the implementation was in the `systemd-base.py` but the template itself was `systemd-user.py`. That is the reason for adding the `AddServiceMethod()` API.

As the decorations did not work in this case, but still wanting to use decorations, I added said method, but did *not* expose it via D-Bus (I did not decorate it). Maybe it would be clearer if it used lower case.

dbusmock/templates/systemd-system.py
jadahl: The only difference indeed is system vs session bus. If this could be part of the template instantiation, I could merge the three files into a single one, which would mean the `AddServiceMethod()` can be avoided while still being able to use decorations on the method.

dbusmock/templates/systemd-base.py
jadahl: As mentioned, when the template is `systemd-user.py` but all the implementation is in `systemd-user.py`, I didn't get any D-Bus methods to work using decorations.

martinpitt:
@jadahl : I reworked the API and CLI a bit in commits d535c0f92b6503a9f3, 1cc4402b6fb43, and cb0161a2351e0fa1 (plus two more which are just tests). Now you can override (or set, if not given at all) a template's `SYSTEM_BUS` both in `spawn_server_template(..., system_bus=False/True)` and on the CLI with `--system/--session`.I hope that should be enough to unify the systemd template?

I recommend to write the template with `SYSTEM_BUS=True`, as that still feels like the more common use case. It's IMHO better to pick *some* default than always requiring an explicit selection.

jadahl:
> I hope that should be enough to unify the systemd template?

Seems to be enough. It's now in a single file, the method decorations work without any need for the AddServiceMethod() helper, so dropped that.

dbusmock/templates/systemd-base.py
jadahl: No longer relevant, so resolving.

dbusmock/mockobject.py
```diff
@@ -400,6 +430,42 @@ def AddMethods(self, interface: str, methods: List[MethodType]) -> None:
...
+        # this for container types.
+        if not isinstance(value, (dbus.Dictionary, dbus.Array)):
+            value = copy.copy(value)
+
+        self.props.setdefault(interface, {})[name] = value
+
+    @dbus.service.method(MOCK_IFACE,
+                         in_signature='sa{sv}',
+                         out_signature='')
+    def UpdateProperties(self, interface: str, properties: PropsType) -> None:
```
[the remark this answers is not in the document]
jadahl: Tests added.

dbusmock/mockobject.py
```diff
@@ -88,6 +88,36 @@ def format_arg(a):
...
+        dbus.types.UInt32,
+    ]
+    if isinstance(value, dbus.String):
+        return dbus.String(str(value), variant_level=1)
+    if type(value) in dbus_types:
+        return type(value)(value.conjugate(), variant_level=1)
+    if isinstance(value, str):
+        return dbus.String(value, variant_level=1)
+    raise dbus.exceptions.DBusException(
+        'couldnt wrap type {}'.format(type(value)))
```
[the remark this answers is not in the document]
jadahl: Changed.

dbusmock/mockobject.py
```diff
@@ -400,6 +430,42 @@ def AddMethods(self, interface: str, methods: List[MethodType]) -> None:
...
+        '''
+        changed_props = {}
+
+        for name, value in properties.items():
+            if not interface:
+                interface = self.interface
+            if name not in self.props.get(interface, {}):
+                raise dbus.exceptions.DBusException(
+                    'property %s not found' % name,
+                    name=self.interface + '.NoSuchProperty')
```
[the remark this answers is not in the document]
jadahl: Fixed.

dbusmock/mockobject.py
jadahl: No longer relevant, AddServiceMethod dropped as it was no longer needed.

dbusmock/templates/systemd-base.py
jadahl: It's now `mock`.

tests/test_api.py
```diff
@@ -338,6 +338,37 @@ def test_properties(self):
         self.assertEqual(self.dbus_props.Get('org.freedesktop.Test.Other', 'color'),
                          'yellow')
 
+        changed_props = []
```
martinpitt, line 341: Thanks! Please squash this commit into the previous one that adds the API.

tests/test_api.py
```diff
@@ -338,6 +338,37 @@ def test_properties(self):
...
+        ml = GLib.MainLoop()
+
+        def catch(*args, **kwargs):
+            if kwargs['interface'] == 'org.freedesktop.DBus.Properties' and \
+               kwargs['member'] == 'PropertiesChanged':
```
martinpitt, line 346: Is there any other signal that would/could happen here? I.e. I wonder if these could  be assertions instead of conditions?

tests/test_api.py
```diff
@@ -338,6 +338,37 @@ def test_properties(self):
...
+        })
+
+        GLib.timeout_add(3000, ml.quit)
+        ml.run()
+
+        self.assertEqual(self.dbus_props.GetAll('org.freedesktop.Test.Main'),
+                         {'version': 5, 'connected': False})
+        self.assertEqual(len(changed_props), 1)
+        self.assertEqual(changed_props[0],
+                         {'version': 5, 'connected': False})
```
martinpitt, line 370: Please drop the first and compare the entire `changed_props` to `[{...}]`. With that style, assertion failures are much more useful, as they will actually show you the extra unexpected entries.

dbusmock/templates/systemd.py
```diff
@@ -0,0 +1,106 @@
+'''systemd mock template
```
martinpitt, line 1: Please adjust the commit message -- that still talks about -user and -system templates and the shared -base.

tests/test_systemd.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/python3
```
martinpitt, line 1: Please squash this "add systemd test" commit into the previous one that adds the template.

martinpitt (changes requested):
Very cool, so much simpler! Just some simple final cleanup to do, then this is ready to land.

How soon do you want/need this released? Shall I wait for @3v1n0 's pending PRs for accountsservice and #72 , or do a release right after this PR lands?

Thanks!

tests/test_api.py
jadahl: Squashed.

tests/test_api.py
jadahl: Dropped the length check.

dbusmock/templates/systemd.py
jadahl: Adjusted.

tests/test_api.py
jadahl: There are also `NameOwnerChanged` signals flying by. I changed to assert a bit, but ignore non-Properties signals.

tests/test_systemd.py
jadahl: Squashed.

jadahl:
> How soon do you want/need this released? Shall I wait for @3v1n0 's pending PRs for accountsservice and #72 , or do a release right after this PR lands?

My use case depends on both a systemd template and an accountsservice template, so no need to do a release without that for me.

martinpitt (approved):
LGTM now, thanks a lot! :+1:

dbusmock/mockobject.py
```diff
@@ -88,6 +88,36 @@ def format_arg(a):
...
+        dbus.types.Byte,
+        dbus.types.Double,
+        dbus.types.Int64,
+        dbus.types.String,
+        dbus.types.UInt32,
+    ]
+    if isinstance(value, dbus.String):
+        return dbus.String(str(value), variant_level=1)
+    if type(value) in dbus_types:
+        return type(value)(value.conjugate(), variant_level=1)
```
cschramm, line 114: This cannot work for at least `Array`, `ByteArray`, `ObjectPath`, `Struct`, `Dictionary`, and `Signature`. Am I missing something here? I tried to use the `UpdateProperties` method with an array property and got the expected exception that `value` does not have `conjugate`.

PR #72, opened 2021-03-22
mockobject: Treat dbus methods as they are without redefining them

3v1n0:
When a dbus method (i.e. a function with @dbus.service.method decorator)
is passed to the function, we are currently redefining a wrapper dbus
method that calls it again.

While this works ok for basic operations, it may make impossible to use
further advanced techniques.

So, just reuse the method as it is when possible.

In particular, this will fix these cases:
 - AddMethod can be called with invalid (even None) arguments signatures
   if the passed function is an actual dbus method
 - Arguments names will be exposed to the introspection using meaningful
   values, based on the actual mock functions arguments names
 - It will be possible again to use all the decorator parameters (such
   as dbus_sender_keyword, dbus_path_keyword or async_callbacks) that
   can be used for some advanced mocking.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -361,7 +361,6 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
 
         if not interface:
             interface = self.interface
-        n_args = len(dbus.Signature(in_sig))
 
         # we need to have separate methods for dbus-python, so clone
         # mock_method(); using message_keyword with this dynamic approach fails
@@ -370,13 +369,20 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         method = lambda self, *args, **kwargs: DBusMockObject.mock_method(
             self, interface, name, in_sig, *args, **kwargs)
 
-        # we cannot specify in_signature here, as that trips over a consistency
-        # check in dbus-python; we need to set it manually instead
-        dbus_method = dbus.service.method(interface,
-                                          out_signature=out_sig)(method)
-        dbus_method.__name__ = str(name)
-        dbus_method._dbus_in_signature = in_sig
-        dbus_method._dbus_args = ['arg%i' % i for i in range(1, n_args + 1)]
+        if isinstance(code, types.FunctionType) and hasattr(code, '_dbus_args'):
+            dbus_method = code
+            in_sig = dbus_method._dbus_in_signature  # type: ignore
+            out_sig = dbus_method._dbus_out_signature  # type: ignore
+        else:
+            # we cannot specify in_signature here, as that trips over a consistency
+            # check in dbus-python; we need to set it manually instead
+            n_args = len(dbus.Signature(in_sig))
+            dbus_args = ['arg%i' % i for i in range(1, n_args + 1)]
+            dbus_method = dbus.service.method(interface,
+                                              out_signature=out_sig)(method)
+            dbus_method.__name__ = str(name)
+            dbus_method._dbus_in_signature = in_sig  # type: ignore
+            dbus_method._dbus_args = dbus_args  # type: ignore
 
         # for convenience, add mocked methods on the primary interface as
         # callable methods


martinpitt (changes requested):
Thanks! It's not obvious to me why you would do that -- the purpose of `AddMethod()` is precisely to *not* have to define a full-blown decorated function. That is just a shortcut for dynamically defining the latter. It also doesn't work over D-Bus, as you can just pass strings there, not functions. Finally, the `code` argument is typed as `str`, so trying to pass a function there would upset the type checker.

If there is a good use case to mix this, then please write a test_api.py unit test for it that illustrates it, and update the type annotation and docstring of `AddMethod`.

This is the same complaint that I had [in the accountsservice PR](https://github.com/martinpitt/python-dbusmock/pull/68#discussion_r598148089). My gut feeling is that you are trying to do something with "properly" defined decorated methods that doesn't work as expected -- can you please tell me what it is, and then we try and fix that properly?

(Finally: I'm quite sure the commit subject meant to say "treat", not "threat" :grin: )

3v1n0:
> Thanks! It's not obvious to me why you would do that -- the purpose of `AddMethod()` is precisely to _not_ have to define a full-blown decorated function. That is just a shortcut for dynamically defining the latter. It also doesn't work over D-Bus, as you can just pass strings there, not functions.

Yeah, this wasn't indeed expected to work under DBus, and yes that's true that code is expected to be a string, but then in `mock_method` we actually consider the fact of code being actually a function type, so while isn't exposed as an API it's actually supported when AddMethod is called from python.

So my point here was to make it possible to avoid re-defining the arguments type when code is a function (maybe via another method) when it's used via actual python code. Making it clear that we support it and if we do, to do it properly.

> Finally, the `code` argument is typed as `str`, so trying to pass a function there would upset the type checker.

Oh, right... I just decided to mute the checker initially as I wanted this for comments first.

> If there is a good use case to mix this, then please write a test_api.py unit test for it that illustrates it, and update the type annotation and docstring of `AddMethod`.
> This is the same complaint that I had [in the accountsservice PR](https://github.com/martinpitt/python-dbusmock/pull/68#discussion_r598148089). My gut feeling is that you are trying to do something with "properly" defined decorated methods that doesn't work as expected -- can you please tell me what it is, and then we try and fix that properly?

So... My point was able to use async callbacks in order to be able something like this https://github.com/3v1n0/python-dbusmock/commit/dac4db16708a4dc8d52b9912bbc327c9840ed8ea in the systemd templates.

And not explicitly depending on GLib directly there, just using the python DBus API and potentially allow to do something similar to what we do in the polkitd template (do a call, make it wait, and explicitly make it return it with some delay).

Other than addressing the issue you mentioned in the accountsservice PR: avoid creating another method on top of something that is already a method on various mocks.

> (Finally: I'm quite sure the commit subject meant to say "treat", not "threat" grin )

LOL, indeed :-)

martinpitt:
> So my point here was to make it possible to avoid re-defining the arguments type when code is a function

Sorry, I still don't see the problem -- it's all fine and well to define functions properly, with decorators, async, etc. -- but if you do that, then why do you need to call `AddMethod()` in addition? That will define them *twice* -- it really should be one **or** the other. So I guess I'm missing something here? What is broken if you simply drop the extra AddMethod call?

3v1n0:
> > So my point here was to make it possible to avoid re-defining the arguments type when code is a function
> 
> Sorry, I still don't see the problem -- it's all fine and well to define functions properly, with decorators, async, etc. -- but if you do that, then why do you need to call `AddMethod()` in addition? That will define them _twice_ -- it really should be one **or** the other. So I guess I'm missing something here? What is broken if you simply drop the extra AddMethod call?

The problem was present in the systemd mock before Jonas refactored it based on your proposal, when we were using a base class loader, that forced us to call the `AddMethod` functions or they were not visible IIRC.

I think for some reasons i had also this issue with accountsservice, but looks like I was doing something wrong as it's now working fine.

martinpitt:
Right, so I guess that was more or less a misunderstanding, and we don't need this?

3v1n0:
> Right, so I guess that was more or less a misunderstanding, and we don't need this?

I think not super needed at this point, I'm wondering if we could then maybe just make `AddMethod` to be stronger on having a string typed code

martinpitt:
Closing for now. Not required in its current form, and we can reopen this if you have another idea for improving this. Thanks!

Issue #74, opened 2021-06-14
Tests fail with Python 3.10


hadess:
Trying to build (and test) python-dbusmock-0.23 in Fedora rawhide against Python 3.10:
```
======================================================================
FAIL: test_methods_type_mismatch (tests.test_api.TestAPI)
calling methods with wrong arguments
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-dbusmock-0.23.0/tests/test_api.py", line 189, in check
    self.dbus_test.Do(*args)
  File "/usr/lib64/python3.10/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.TypeError: Traceback (most recent call last):
  File "/usr/lib64/python3.10/site-packages/dbus/service.py", line 711, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/builddir/build/BUILD/python-dbusmock-0.23.0/dbusmock/mockobject.py", line 400, in <lambda>
    method = lambda self, *args, **kwargs: DBusMockObject.mock_method(
  File "/builddir/build/BUILD/python-dbusmock-0.23.0/dbusmock/mockobject.py", line 661, in mock_method
    m.append(signature=in_signature, *args)
TypeError: 'dbus.String' object cannot be interpreted as an integer
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-dbusmock-0.23.0/tests/test_api.py", line 205, in test_methods_type_mismatch
    check('i', ['hello'], 'TypeError: an integer is required')
  File "/builddir/build/BUILD/python-dbusmock-0.23.0/tests/test_api.py", line 193, in check
    self.assertIn(err, str(e))
[...]

Issue #75, opened 2021-06-14
AccountServices mock test hangs


hadess:
For reasons I don't understand, python-dbusmock 0.23's test fails when run in a terminal on my desktop machine.

```
+ /usr/bin/python3 -m unittest -v
test_automatic_login_users (tests.test_accounts_service.TestAccountsService) ... 
(process:588754): accountsservice-WARNING **: 11:56:38.691: ActUserManager: user (null) has no username (uid: -1)
[...]
```

Given that python-dbusmock doesn't use meson or another build system with builtin timeouts, `wait_changed()` should really be modified to have an upper-bound so it doesn't hang.

hadess:
in `tearDown(self)`:
```patch
-        while self.p_manager.props.is_loaded:
-            self.ctx.iteration(True)
```
Fixes it.

I'm not sure that the accountsservice library handles the daemon not being running properly as that would be from the non-mocked daemon being auto-started by D-Bus:
```
(process:589209): accountsservice-WARNING **: 12:04:29.883: ActUserManager: user (null) has no username (uid: -1)
```

hadess:
CC @3v1n0 as he wrote that code in #68

hadess:
Using python3-3.9.5-2.fc34.x86_64 and accountsservice-libs-0.6.55-6.fc34.x86_64

hadess:
> Fixes it.

It fixes the hang (if one also removes the assertion just below in tearDown), but other tests still fail:
```
======================================================================
ERROR: test_automatic_login_users_via_mock (__main__.TestAccountsService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/hadess/Projects/Fedora/merged/python-dbusmock/python-dbusmock-0.23.0/./tests/test_accounts_service.py", line 312, in test_automatic_login_users_via_mock
    user = self.p_manager.create_user(
gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.2 was not provided by any .service files (2)
```

whot:
There's something else weird with these tests. Any instantiation of the user manager object causes a SIGTERM shortly after the accounts service tests finish:
```
pytest -vv
============================================================================================================================= test session starts ==============================================================================================================================
platform linux -- Python 3.10.4, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /usr/bin/python3
cachedir: .pytest_cache
[...]

PR #76, opened 2021-06-14
Work-around error text changes

hadess:
https://github.com/martinpitt/python-dbusmock/issues/74

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -202,7 +202,8 @@ assert args[2] == 5
 
         # type mismatch
         check('u', [-1], 'convert negative value to unsigned')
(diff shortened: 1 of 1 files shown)


hadess:
Worked with Python 3.10 in rawhide. I'm guessing that it'll break for anything older. So probably incorrect, but that's the fix that'll be in rawhide so I can get a python-dbusmock built against Python 3.10 to make it installable, and unblock another upstream CI.

martinpitt:
Thanks! I did something similar in commit 1a8cd4c5a2a93b which works everywhere. I'll do a new point release now, as issue #75 seems to require some  more time to investigate.

PR #78, opened 2021-06-25
logind: Implement Inhibit and ListInhibitors

benzea:
Properly track held inhibitors, replacing the previous dummy
implementation.

Note that previously the FD represented by the passed Inhibit_fd integer
would be passed back. This was never useful though, as it is impossible
to do anything interesting by setting the FD.

### DIFF
--- a/dbusmock/templates/logind.py
+++ b/dbusmock/templates/logind.py
@@ -17,6 +17,7 @@ __copyright__ = '(c) 2013 Canonical Ltd.'
 import os
 import dbus
 
+from gi.repository import GLib
 from dbusmock import MOCK_IFACE, mockobject
 
 BUS_NAME = 'org.freedesktop.login1'
@@ -58,9 +59,6 @@ def load(mock, parameters):
         ('GetUser', 'u', 'o', 'ret = "/org/freedesktop/login1/user/%u" % args[0]'),
         ('KillUser', 'us', '', ''),
         ('TerminateUser', 'u', '', ''),
-
-        ('Inhibit', 'ssss', 'h', 'ret = %i' % parameters.get('Inhibit_fd', 3)),
-        ('ListInhibitors', '', 'a(ssssuu)', 'ret = []'),
     ])
 
     mock.AddProperties(MAIN_IFACE,
@@ -104,6 +102,41 @@ def ListSessions(_):
             seat = obj.Get('org.freedesktop.login1.Session', 'Seat')[0]
             sessions.append((session_id, uid, username, seat, k))
     return sessions
+
+
+@dbus.service.method(MAIN_IFACE,
+                     in_signature='ssss',
+                     out_signature='h')
+def Inhibit(_, what, who, why, mode):
+    if not hasattr(mockobject, "inhibitors"):
+        mockobject.inhibitors = list()
+
+    fd_r, fd_w = os.pipe()
+
+    inhibitor = (what, who, why, mode, 1000, 123456)
+    mockobject.inhibitors.append(inhibitor)
+
+    def inhibitor_dropped(fd, cond):
+        # pylint: disable=unused-argument
+        os.close(fd)
+        mockobject.inhibitors.remove(inhibitor)
+        return False
+
+    GLib.unix_fd_add_full(GLib.PRIORITY_HIGH, fd_r, GLib.IO_HUP, inhibitor_dropped)
+    GLib.idle_add(os.close, fd_w)
+
+    return fd_w
+
+
+@dbus.service.method(MAIN_IFACE,
+                     in_signature='',
+                     out_signature='a(ssssuu)')
+def ListInhibitors(_):
+    if not hasattr(mockobject, "inhibitors"):
+        mockobject.inhibitors = list()
+
+    return mockobject.inhibitors
+
 
 #
 # Convenience methods on the mock
--- a/tests/test_logind.py
+++ b/tests/test_logind.py
@@ -113,6 +113,23 @@ class TestLogind(dbusmock.DBusTestCase):
         self.assertEqual(props['PreparingForSleep'], False)
         self.assertEqual(props['IdleSinceHint'], 0)
 
+    def test_inhibit(self):
+        (self.p_mock, obj_logind) = self.spawn_server_template('logind', {}, stdout=subprocess.PIPE)
+
+        # what, who, why, mode
+        fd = obj_logind.Inhibit('suspend', 'testcode', 'purpose', 'delay')
+
+        # Our inhibitor is held
+        out = subprocess.check_output(['systemd-inhibit'],
+                                      universal_newlines=True)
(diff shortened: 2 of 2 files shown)


benzea:
CC: @3v1n0

martinpitt:
Adding my two cents, to me this feels a bit over the top -- this is not a mock any more, but an actual reimplementation of systemd. The idea of a mock is that a test case modifies e.g. `ListInhibitors()` at some adequate time to return a static result. This keeps unit tests explicit.

I don't veto this if you need this, but I'm curious about the use case?

benzea:
My intention was being able to test that the inhibitor is correctly released by the application.

martinpitt:
OK. I'm a bit wary of existing consumers of this template, as this changes existing behaviour. I. e. in this case it may be better to put this into a local mock, or enable this new behaviour with a template parameter? The template already accepts `Inhibit_fd`, it could look for value `dynamic` or a new parameter `dynamic_inhibit: true`?

DylanVanAssche:
This PR would benefit me as well for https://gitlab.freedesktop.org/upower/upower/-/merge_requests/67#note_1001446

For UPower, an inhibitor lock is taken and released for some operations.
It would be great to have this patch to test if the lock is taken and released properly.

benzea:
If you need something simpler or in the short term, then this code from fprintd might be an option:

  https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/feature/suspend-resume/tests/fprintd.py#L588

Not sure whether merging something here or having this kind of thing in each test-suite is better.

DylanVanAssche:
@benzea I prefer upstream if possible :) 
However, I have no strong opinion on this, I mostly need this because @hadess asked for it.

hadess:
>  However, I have no strong opinion on this, I mostly need this because @hadess asked for it.

Have you checked that this actually helps with the implementation of the test you wanted? If it's enough, it's likely that just making `ListInhibitors` return a non-empty list if `Inhibit` was called would be enough for the tests.

Most of the complexity in the code is because of the FD handling. If we can hide that, it's actually quite trivial, just updating a list depending on whether a fd is opened or got closed...

DylanVanAssche:
> > However, I have no strong opinion on this, I mostly need this because @hadess asked for it.
> 
> Have you checked that this actually helps with the implementation of the test you wanted? If it's enough, it's likely that just making `ListInhibitors` return a non-empty list if `Inhibit` was called would be enough for the tests.
> 
> Most of the complexity in the code is because of the FD handling. If we can hide that, it's actually quite trivial, just updating a list depending on whether a fd is opened or got closed...

The test should only do this:

- Let it take an inhibitor lock
- Verify that the list is not empty
- Drop inhibitor
- Verify that the list is empty

So yes, that's why I thought this PR would help here, without reinventing the wheel.
I looked at the code and it looks like it does exactly this:

- Take inhibitor lock: open fd and add to list
- Drop inhibitor lock: close fd and remove from list

I just don't want to do *any* work anymore which will not be accepted in the end.

hadess:
> The template already accepts `Inhibit_fd`, it could look for value `dynamic` or a new parameter `dynamic_inhibit: true`?

@benzea do you want to look at doing that, or shall I?

benzea:
I think the `Inhibit_fd` parameter has never been usable and really should be removed.

If we want to simplify this, I would be happy to just remove the `pid`/`uid` resolving part. Not like that is really useful during testing. I don't think that keeping the list of inhibitors actually makes this code much more complex, considering we need to create at least one pipe and track closing it anyway.

hadess:
> If we want to simplify this, I would be happy to just remove the `pid`/`uid` resolving part.


> Not like that is really useful during testing. I don't think that keeping the list of inhibitors actually makes this code much more complex, considering we need to create at least one pipe and track closing it anyway.

It's complex compared to having a Get/Set call on the mock interface for the inhibitors variable, for example.

benzea:
> It's complex compared to having a Get/Set call on the mock interface for the inhibitors variable, for example.

Sure, though my approach doesn't actually need any API on the mock interface. So, I think one can argue either way with regard to complexity …

At the end of the day, I don't care. I probably won't even use this functionality for a long time as I have a working solution for the fprintd CI as of now.

martinpitt:
OK, @benzea 's argument that the current fd is not useful for anything is convincing, I just don't like to break the existing API. (Again: for issues like these it's much easier to modify the mock dynamically on the fly, or keep the one in your tests which exactly matches your needs)

I am admittedly a bit confused: Who wants/needs this? @benzea and @DylanVanAssche ? Is the current implementation good for both of you, i.e. should we land this approach?

Note that `test_inhibit` currently fails everywhere.

benzea:
@martinpitt, I believe this covers all possible use cases:

 1. If grabbing/dropping an inhibitor should be tested, one can poll `ListInhibitors` to track this
 2. If `ListInhibitors` needs to be mocked, then one can use the usual `Inhibit` API

Oops, the test failure was because I removed the uid/pid resolution, systemd-inhibit couldn't lookup an executable, replacing it with `N/A` which results in an extra space. Updated the RE to use ` +` so that space padding in the output is fine.

benzea:
To be fair, emulating ListInhibitors is probably not very useful. One would likely want to emulate some of the properties instead (which can be controlled differently already).

martinpitt:
Still a [pylint error](https://github.com/martinpitt/python-dbusmock/pull/78/checks?check_run_id=3209391266)
```
************* Module dbusmock.templates.logind
dbusmock/templates/logind.py:112:8: W0212: Access to a protected member _inhibitors of a client class (protected-access)
dbusmock/templates/logind.py:117:4: W0212: Access to a protected member _inhibitors of a client class (protected-access)
dbusmock/templates/logind.py:122:8: W0212: Access to a protected member _inhibitors of a client class (protected-access)
dbusmock/templates/logind.py:136:8: W0212: Access to a protected member _inhibitors of a client class (protected-access)
dbusmock/templates/logind.py:138:11: W0212: Access to a protected member _inhibitors of a client class (protected-access)
test_pylint (tests.test_code.StaticCodeTests) ... ERROR
```

martinpitt (approved):
Thanks @benzea ! As I'm still not quite clear who should have the final word about this (i.e. who uses this), please land if/when ready.

DylanVanAssche:
@martinpitt These changes seem perfect for my use case in UPower!
I tested this change here: https://gitlab.freedesktop.org/dylanvanassche/upower/-/merge_requests/1/diffs#630c35909d84d2320a6a948b657cd2c6b66a31b8_982_1005
and it seems working great!

The `ListInhibitors` lists correctly that we have an inhibitor lock and that the lock is released afterwards.

LGTM! Thanks a lot for pushing this forward @benzea & @martinpitt

benzea:
@martinpitt, lets merge this then. Would do it myself, but don't have the permission.

PR #79, opened 2021-07-19
templates: Add power-profiles-daemon template

hadess:
My Python-fu is weak, so I couldn't figure out how to append to the `ActiveProfileHolds` property (which is supposed to be an array), or how I could reference that recently appended "hold" to a separate store for `ReleaseHold` to use.

This is the error I get when I try to use the `.append()` path:
```
ERROR:dbus.service:Unable to append (dbus.Array([{'Profile': dbus.String('power-saver'), 'Reason': dbus.String('"g-s-d mock test"'), 'ApplicationId': dbus.String('org.gnome.SettingsDaemon.Power')}], signature=dbus.Signature('(aa{sv})')),) to message with signature v: <class 'TypeError'>: string indices must be integers
  File "/home/hadess/Projects/gnome-install/bin/powerprofilesctl", line 124, in get_profiles_property
    profiles = proxy.Get('(ss)', 'net.hadess.PowerProfiles', prop)
  File "/usr/lib/python3.9/site-packages/gi/overrides/Gio.py", line 349, in __call__
    result = self.dbus_proxy.call_sync(self.method_name, arg_variant,
gi.repository.GLib.Error: g-io-error-quark: GDBus.Error:org.freedesktop.DBus.Python.TypeError: Traceback (most recent call last):
  File "/usr/lib64/python3.9/site-packages/dbus/service.py", line 755, in _message_cb
    _method_reply_return(connection, message, method_name, signature, *retval)
  File "/usr/lib64/python3.9/site-packages/dbus/service.py", line 256, in _method_reply_return
    reply.append(signature=signature, *retval)
TypeError: string indices must be integers
 (36)
```

### DIFF
--- a/tests/run-fedora
+++ b/tests/run-fedora
@@ -1,6 +1,11 @@
 #!/bin/sh
 set -eux
 # install build dependencies
+source /etc/os-release
+if [ $VERSION_ID == 34 ]; then
+    dnf -y install dnf-plugins-core
+    dnf -y copr enable hadess/power-profiles-daemon # FIXME remove when using Fedora 35 or newer
+fi
 dnf -y install python3-setuptools python3-gobject-base \
     python3-dbus dbus-x11 python3-pycodestyle python3-pyflakes python3-pylint python3-pip \
     upower NetworkManager bluez libnotify polkit power-profiles-daemon
--- a/tests/run-fedora
+++ b/tests/run-fedora
@@ -3,7 +3,7 @@ set -eux
 # install build dependencies
 dnf -y install python3-setuptools python3-gobject-base \
     python3-dbus dbus-x11 python3-pycodestyle python3-pyflakes python3-pylint python3-pip \
-    upower NetworkManager bluez libnotify polkit
+    upower NetworkManager bluez libnotify polkit power-profiles-daemon
 
 # systemd's tools otherwise fail on "not been booted with systemd"
 mkdir -p /run/systemd/system
--- a/tests/test_power_profiles_daemon.py
+++ b/tests/test_power_profiles_daemon.py
@@ -1,0 +1,148 @@
+#!/usr/bin/python3
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Bastien Nocera'
+__copyright__ = '(c) 2021 Red Hat Inc.'
+
+import fcntl
+import os
+import shutil
+import subprocess
+import sys
+import unittest
+import time
+
+import dbus
+import dbus.mainloop.glib
+
+import dbusmock
+
+dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+
+have_powerprofilesctl = shutil.which('powerprofilesctl')
+
+
+@unittest.skipUnless(have_powerprofilesctl, 'powerprofilesctl not installed')
+class TestPowerProfilesDaemon(dbusmock.DBusTestCase):
+    '''Test mocking power-profiles-daemon'''
+    @classmethod
+    def setUpClass(cls):
+        cls.start_system_bus()
+        cls.dbus_con = cls.get_dbus(True)
+
+    def setUp(self):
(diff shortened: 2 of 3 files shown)


hadess:
I've figured out how to add elements to the property. Will try to implement release now.

hadess:
Array handling was pretty painful, but managed to get it sorted. The test suite will likely fail while the version of power-profiles-daemon is too old on Fedora, but should run locally with the current git main of power-profiles-daemon.

I'm finishing implementing [this gnome-settings-daemon feature](https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/merge_requests/247) (and writing a test using this new template) before doing a release.

hadess:
Tested and works. This just needs a newer power-profiles-daemon available for the test suite to work.

hadess:
The new power-profiles-daemon with the `powerprofilesctl` with all the right bits is now available in Fedora Rawhide:
https://koji.fedoraproject.org/koji/buildinfo?buildID=1782223

hadess:
Had to create a COPR to get the latest version of power-profiles-daemon in F34:
https://copr.fedorainfracloud.org/coprs/hadess/power-profiles-daemon/

Let me know whether there's anything else!

tests/test_power_profiles_daemon.py
```diff
@@ -0,0 +1,148 @@
...
+                                'sleep', '60'],
+                               stdout=subprocess.PIPE)
+        time.sleep(0.3)
+
+        out = self.run_powerprofilesctl_list_holds()
+        self.assertEqual(out, 'Hold:\n'
+                              '  Profile:         power-saver\n'
+                              '  Application ID:'
+                              '  org.gnome.SettingsDaemon.Power\n'
+                              '  Reason:          g-s-d mock test\n')
```
martinpitt, line 93: Hint for the future: Multi-line strings with `'''` or `"""` are a bit nicer for this.

tests/test_power_profiles_daemon.py
```diff
@@ -0,0 +1,148 @@
...
+        out = self.run_powerprofilesctl_list_holds()
+        self.assertEqual(out, '')
+
+        # 1 hold
+        cmd = subprocess.Popen(['powerprofilesctl', 'launch', '-p',
+                                'power-saver', '-r', 'g-s-d mock test',
+                                '-i', 'org.gnome.SettingsDaemon.Power',
+                                'sleep', '60'],
+                               stdout=subprocess.PIPE)
+        time.sleep(0.3)
```
martinpitt, line 86: Meh, but as long as this is enough for tests, sure. (Usually retry loops are better)

tests/run-fedora
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 set -eux
 # install build dependencies
+source /etc/os-release
+if [ $VERSION_ID == 34 ]; then
+    dnf -y install dnf-plugins-core
+    dnf -y copr enable hadess/power-profiles-daemon # FIXME remove when using Fedora 35 or newer
+fi
```
martinpitt, line 8: That's actually nice! (As an alternative, it would have been enough to run this for Rawhide only, and not install it on 34).

martinpitt (approved):
Merci beaucoup ! This is pretty much your tool anyway, so not commenting too much about whether this is sufficiently powerful/useful as a general template.

tests/test_power_profiles_daemon.py
hadess: I forgot about this, but this would actually make the string less readable, as we'd need to nuke the indentation, and rely on implicit linefeeds.

hadess:
Thanks for the merge, but squashing the commits used the tests commit's commit message...

tests/test_power_profiles_daemon.py
hadess: Fixed in https://github.com/martinpitt/python-dbusmock/pull/80

martinpitt:
@hadess : WDYM about the commit message? https://github.com/martinpitt/python-dbusmock/commit/9d0be456193289 seems ok? I did keep the part about requiring the 0.9 version as that seems useful.

hadess:
> @hadess : WDYM about the commit message? [9d0be45](https://github.com/martinpitt/python-dbusmock/commit/9d0be456193289) seems ok? I did keep the part about requiring the 0.9 version as that seems useful.

You're right, not sure how I read that incorrectly.

PR #80, opened 2021-07-25
tests: Fix static sleep in power-profile-daemons test

hadess:
See https://github.com/martinpitt/python-dbusmock/pull/79#discussion_r676145708


martinpitt (approved):
Merci !

PR #83, opened 2021-08-11
templates: Add IIO Sensors Proxy support

3v1n0:
Add template to run a mock IIO Sensors Proxy, supporting accelerometers,
compasses, proximity and light sensors.

Includes unit tests, both checking the dbus interface directly and using
the monitor-sensor debug tool, if available.

This was initially part of [this mutter MR](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1233), but @hadess suggested to upstream it for usage in other components such as g-c-c.

### DIFF
--- a/dbusmock/templates/iio-sensors-proxy.py
+++ b/dbusmock/templates/iio-sensors-proxy.py
@@ -1,0 +1,204 @@
+'''sensors proxy mock template
+'''
+
+# This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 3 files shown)


3v1n0:
@martinpitt hey, need something else here?

martinpitt (approved):
Whoops, sorry @3v1n0, this slipped through the cracks. Thanks!

PR #84, opened 2021-08-27
dbusmock: Disallow service activation on system bus

benzea:
Even with standard_system_servicedirs empty, dbus-daemon will always add
the standard search directories in order to activate services. However,
in a test environment it is unexpected that a system service might be
started.

As such, disallow all calls to StartServiceByName, preventing any such
situation.

### DIFF
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -60,7 +60,7 @@ class DBusTestCase(unittest.TestCase):
   <type>system</type>
   <keep_umask/>
   <listen>unix:tmpdir=/tmp</listen>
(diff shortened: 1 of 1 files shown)


martinpitt:
@benzea : FYI, I just pushed fixes for the new pylint warnings that broke the Fedora tests. So if you figure out a solution here, please rebase. Thanks!

benzea:
Hah, much easier, I just didn't understand how the configuration works.

martinpitt (approved):
Hah, excellent, thank you! I actually wonder how it survived like that for that long!

PR #87, opened 2021-10-27
cli: Add support for running custom commands or shell on the mock environment

3v1n0:
It may be used to quickly debug interfaces

Allowing to use something like `python3 -m dbusmock -t upower -c d-feet`

### DIFF
--- a/dbusmock/__main__.py
+++ b/dbusmock/__main__.py
@@ -15,6 +15,8 @@ __copyright__ = '''
 
 import argparse
 import json
+import os
+import subprocess
 import sys
 
 import dbusmock.mockobject
@@ -43,6 +45,8 @@ def parse_args():
                         help='automatically implement the org.freedesktop.DBus.ObjectManager interface')
     parser.add_argument('-p', '--parameters',
                         help='JSON dictionary of parameters to pass to the template')
+    parser.add_argument('-e', '--exec', nargs=argparse.REMAINDER,
+                        help='Command to run in the mock environment')
 
     arguments = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


3v1n0:
Rebased to fully use `ctypes` instead of `gi.repository`

dbusmock/__main__.py
```diff
@@ -43,6 +45,10 @@ def parse_args():
                         help='automatically implement the org.freedesktop.DBus.ObjectManager interface')
     parser.add_argument('-p', '--parameters',
                         help='JSON dictionary of parameters to pass to the template')
+    parser.add_argument('-x', '--shell', action='store_true',
+                        help='Open a SHELL with the given mock for debugging purposes')
+    parser.add_argument('-e', '--exec', nargs=argparse.REMAINDER,
+                        help='Command to run in the mock environment')
```
martinpitt, line 51: I find this too redundant. Surely you could just do `-e bash` to get a shell? That's pretty much what the code does anyway.

dbusmock/__main__.py
```diff
@@ -119,8 +128,33 @@ def parse_args():
...
+            @ctypes.CFUNCTYPE(ctypes.c_bool)
+            def check_process():
+                """ Check if the launched process is still alive """
+
+                if cmd_proc.poll() is None:
+                    return True
+                should_run.pop()
+                return False
+
+            libglib.g_idle_add(check_process)
```
martinpitt, line 151: This is rather "meh" -- would't it be easier to use [unix_signal_add()](https://docs.gtk.org/glib/func.unix_signal_add.html) to listen for SIGCHILD?

martinpitt (changes requested):
This is pretty nice, thank you! Can you please add a section about it to README?

dbusmock/__main__.py
3v1n0: That's fine... It was just a convenience command, I can drop it.

dbusmock/__main__.py
3v1n0: It would, except that it doesn't support it :(

3v1n0: `(process:325649): GLib-CRITICAL **: 23:11:45.387: g_unix_signal_source_new: assertion 'signum == SIGHUP || signum == SIGINT || signum == SIGTERM || signum == SIGUSR1 || signum == SIGUSR2 || signum == SIGWINCH' failed`

martinpitt: Ah right, glib does its own thing with watching processes.. https://docs.gtk.org/glib/func.child_watch_add.html ?

3v1n0: Using that now, then!

martinpitt (approved):
Sorry for letting this slip. This is really nice, thank you!

PR #88, opened 2021-11-04
power_profiles_daemon: Update signature for Profiles array

hadess:
It's declared as an array of a{sv} in the D-Bus interface, but the lack
of annotation means that it's exported as a{ss} instead.


martinpitt (approved):
Oops, merci beaucoup !

PR #89, opened 2021-11-23
bluez5 template fixes

hadess:
Tested with gnome-bluetooth's test suite.

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -111,6 +111,9 @@ def AddAdapter(self, device_name, system_name):
     Returns the new object path.
     '''
     path = '/org/bluez/' + device_name
+    address_start = int(device_name[-1])
+    address = f"{address_start:02d}:{address_start+1:02d}:{address_start+2:02d}:" + \
+        f"{address_start+3:02d}:{address_start+4:02d}:{address_start+5:02d}"
     adapter_properties = {
         'UUIDs': dbus.Array([
             # Reference:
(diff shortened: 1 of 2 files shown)


hadess:
I've removed the ManagedObjects change as I couldn't figure out how the code styling tool wanted me to rework that code.

dbusmock/templates/bluez5.py
```diff
@@ -111,6 +111,9 @@ def AddAdapter(self, device_name, system_name):
     Returns the new object path.
     '''
     path = '/org/bluez/' + device_name
+    address_start = int(device_name[-1])
+    address = f"{address_start:02d}:{address_start+1:02d}:{address_start+2:02d}:" + \
+        f"{address_start+3:02d}:{address_start+4:02d}:{address_start+5:02d}"
```
martinpitt, line 116: This would be cleaner in mod 256 (aka `& 0xFF`), but let's hope that device names are ASCII :grin:

dbusmock/templates/bluez5.py
```diff
@@ -176,6 +176,26 @@ def AddAdapter(self, device_name, system_name):
     return path
 
 
+@dbus.service.method(BLUEZ_MOCK_IFACE,
+                     in_signature='s')
+def RemoveAdapter(self, device_name):
+    '''Convenience method to add a Bluetooth adapter
```
martinpitt, line 182: s/add/remove/

martinpitt (changes requested):
Thanks for the fixes! Just one copy&paste error here. The possible byte overflow is fine as-is IMHO.

dbusmock/templates/bluez5.py
hadess: I really didn't bother with any argument checking, the code above doesn't even check for a valid device_name that could also be a valid D-Bus path object component. Let me know if you want me to add more checks here.

martinpitt (approved):
Merci beaucoup !

PR #93, opened 2021-12-01
mockobject: Wrap dbus.types.Array as is when changing properties

jadahl:
This allows changing property with array as type. It's up to the caller
to make sure the array is populated properly. Not having this results in

    AttributeError: 'dbus.Array' object has no attribute 'conjugate'

when trying to update array properties.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -92,7 +92,6 @@ def _format_args(args):
 
 def _wrap_in_dbus_variant(value):
     dbus_types = [
-        dbus.types.Array,
         dbus.types.ByteArray,
         dbus.types.Int16,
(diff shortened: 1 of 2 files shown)


martinpitt:
Thanks! This should get a unit test to reproduce the original bug and test the fix. Do you want to look at this, or shall I?

jadahl:
> Thanks! This should get a unit test to reproduce the original bug and test the fix. Do you want to look at this, or shall I?

Now tested. Without the first commit, it'd fail with:

```
AttributeError: 'dbus.Array' object has no attribute 'conjugate'
```

martinpitt (approved):
Very nice, thanks a lot!

Issue #92, opened 2021-12-01
Traceback in mock implementation doesn't fail test


hadess:
I've just realised that some tests which I thought passed without a hitch actually caused tracebacks in the mock implementation, most notably things like this:
```
1638374424.085 GetManagedObjects
1638374425.834 RegisterAgent "/org/gnome/bluetooth/integration_test" "DisplayYesNo"
Agent registration failed: GDBus.Error:org.freedesktop.DBus.Python.AttributeError: Traceback (most recent call last):
  File "/usr/lib64/python3.10/site-packages/dbus/service.py", line 715, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/home/hadess/Projects/gnome-install/lib/python3.10/site-packages/dbusmock/mockobject.py", line 398, in <lambda>
    method = lambda self, *args, **kwargs: DBusMockObject.mock_method(
  File "/home/hadess/Projects/gnome-install/lib/python3.10/site-packages/dbusmock/mockobject.py", line 667, in mock_method
    return code(self, *args)
  File "/home/hadess/Projects/gnome-install/lib/python3.10/site-packages/dbusmock/templates/bluez5.py", line 49, in RegisterAgent
    if agent_path in manager.agent_paths:
AttributeError: 'DBusMockObject' object has no attribute 'agent_paths'
```
which doesn't seem to cause an immediate test failure.

The test, in 2 parts, with the client running in a separate process:
https://gitlab.gnome.org/GNOME/gnome-bluetooth/-/blob/master/tests/integration-test#L341
https://gitlab.gnome.org/GNOME/gnome-bluetooth/-/blob/master/tests/integration-test#L259

I don't know whether it's something wrong with my own code, or something which should be extended and checked for in python-dbusmock.

martinpitt:
#94 fixes that general problem. At first sight, the reason here is that dbus-python usually intercepts tracebacks in D-Bus calls  and propagates them to the caller as error. That's a bit irritating indeed. Hopefully that would still fail the test at some point, as the mock didn't return the expected answer?

hadess:
> At first sight, the reason here is that dbus-python usually intercepts tracebacks in D-Bus calls and propagates them to the caller as error.

You're right, it did propagate a D-Bus error which I didn't actually check for. I need to change my local API. Thanks.

PR #94, opened 2021-12-01
More bluez mock implementation fixes

hadess:
See also https://github.com/martinpitt/python-dbusmock/issues/92


martinpitt (approved):
Merci !

PR #95, opened 2021-12-10
logind: Mock TakeControl, TakeDevice and ReleaseDevice

jadahl:
Will help making it possible to run a display server in a mocked
virtual environment, e.g. in a virtual machine.

### DIFF
--- a/dbusmock/templates/logind.py
+++ b/dbusmock/templates/logind.py
@@ -16,6 +16,7 @@ __copyright__ = '(c) 2013 Canonical Ltd.'
 
 import os
 import dbus
+import re
 
 from gi.repository import GLib
 from dbusmock import MOCK_IFACE, mockobject
@@ -23,6 +24,7 @@ from dbusmock import MOCK_IFACE, mockobject
 BUS_NAME = 'org.freedesktop.login1'
 MAIN_OBJ = '/org/freedesktop/login1'
 MAIN_IFACE = 'org.freedesktop.login1.Manager'
+SESSION_IFACE = 'org.freedesktop.login1.Session'
 SYSTEM_BUS = True
 
 
@@ -271,6 +273,7 @@ def AddSession(self, session_id, seat, uid, username, active):
                        ('SetIdleHint', 'b', '', ''),
                        ('Terminate', '', '', ''),
                        ('Unlock', '', '', ''),
+                       ('TakeControl', 'b', '', ''),
                    ])
 
     # add session to seat
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
     obj_user.Set('org.freedesktop.login1.User', 'Sessions', cur_sessions)
 
     return session_path
+
+@dbus.service.method(SESSION_IFACE,
+                     in_signature='uu',
+                     out_signature='hb')
+def TakeDevice(self, major, minor):
+    sysfs_uevent_path = '/sys/dev/char/{}:{}/uevent'.format(major, minor)
+    sysfs_uevent = open(sysfs_uevent_path, 'r')
+    devname = None
+    for line in sysfs_uevent.readlines():
+        match = re.search('DEVNAME=(.*)', line)
+        if match:
+            devname = match[1]
+            break
+    sysfs_uevent.close()
+    if not devname:
+        raise dbus.exceptions.DBusException(f'Device file {major}:{minor} doesn\'t exists',
+                                            major=major, minor=minor)
+    fd = os.open('/dev/' + devname, os.O_RDWR | os.O_CLOEXEC)
+    unix_fd = dbus.types.UnixFd(fd)
+    os.close(fd)
+    return (unix_fd, False)
+
+@dbus.service.method(SESSION_IFACE,
+                     in_signature='uu',
+                     out_signature='')
+def ReleaseDevice(self, major, minor):
+    pass


dbusmock/templates/logind.py
```diff
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
     obj_user.Set('org.freedesktop.login1.User', 'Sessions', cur_sessions)
 
     return session_path
+
```
martinpitt, line 293: Tests fail on pycodestyle:
```
./dbusmock/templates/logind.py:294:1: E302 expected 2 blank lines, found 1
./dbusmock/templates/logind.py:315:1: E302 expected 2 blank lines, found 1
```

dbusmock/templates/logind.py
```diff
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
...
+    sysfs_uevent = open(sysfs_uevent_path, 'r')
+    devname = None
+    for line in sysfs_uevent.readlines():
+        match = re.search('DEVNAME=(.*)', line)
+        if match:
+            devname = match[1]
+            break
+    sysfs_uevent.close()
+    if not devname:
+        raise dbus.exceptions.DBusException(f'Device file {major}:{minor} doesn\'t exists',
```
martinpitt, line 308: Please use `f""` to avoid quoting the `'`, and fix the "exists" typo to "exist"

dbusmock/templates/logind.py
```diff
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
...
+
+@dbus.service.method(SESSION_IFACE,
+                     in_signature='uu',
+                     out_signature='hb')
+def TakeDevice(self, major, minor):
+    sysfs_uevent_path = '/sys/dev/char/{}:{}/uevent'.format(major, minor)
+    sysfs_uevent = open(sysfs_uevent_path, 'r')
+    devname = None
+    for line in sysfs_uevent.readlines():
+        match = re.search('DEVNAME=(.*)', line)
```
martinpitt, line 302: Let's please be correct here and use `re.match`, just in case there is ever a field like "FOODEVNAME".

dbusmock/templates/logind.py
```diff
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
...
+        if match:
+            devname = match[1]
+            break
+    sysfs_uevent.close()
+    if not devname:
+        raise dbus.exceptions.DBusException(f'Device file {major}:{minor} doesn\'t exists',
+                                            major=major, minor=minor)
+    fd = os.open('/dev/' + devname, os.O_RDWR | os.O_CLOEXEC)
```
martinpitt, line 310: Sorry, no, this is not acceptable for a mock. The point of this is specifically to *not* touch the system, and in many cases you will not even be allowed to do this as regular user. This might work in conjunction with https://github.com/martinpitt/umockdev, but then this is so special that I think this template rather belongs into your specific upstream project.

dbusmock/templates/logind.py
```diff
@@ -287,3 +290,30 @@ def AddSession(self, session_id, seat, uid, username, active):
...
+    unix_fd = dbus.types.UnixFd(fd)
+    os.close(fd)
+    return (unix_fd, False)
+
+@dbus.service.method(SESSION_IFACE,
+                     in_signature='uu',
+                     out_signature='')
+def ReleaseDevice(self, major, minor):
+    pass
```
martinpitt, line 319: This is asymmetric -- if TakeDevice() creates an open fd and returns it, then ReleaseDevice should clean it up. I.e., you need to remember the fd inside the mock. 

This comment is just for completeness -- IMHO this is too specific for a generic logind mock.

martinpitt (changes requested):
Thanks! This will need some discussion.

dbusmock/templates/logind.py
martinpitt: What *would* be appropriate for a mock is to drop all the actual /sys and /dev poking, and create/return an fd to /dev/dull or a socketpair() or something, and store it in `self` in a major/minor → fd map. Then clean it up again in ReleaseDevice.

jadahl: I realize this is a bit weird for a mocked logind, but anything less will simply not work in my use case, where the device opened here actually isn't "real" either (it opens a VKMS device node in a QEMU virtual machine, and eventually will open uinput devices too), but still need to open them for real.

Would a "special mode" in logind.py upstream make this acceptable? E.g. by default it would just fail, and if enabled, it'd opn things.  Otherwise I guess I just need to, as you say, copy the whole logind template into where it's used :(

martinpitt: Hiding this behind a parameter is a bit better, but it still feels too special-case-y honestly. It would also need an integration test case to ensure that it actually succeeds opening some device (probably /dev/zero for testing), and implement the reverse action in ReleaseDevice().

There is need to copy the whole logind template, your test can just call `AddMethod()` over D-Bus in your particular test case and poke in a tailored TakeDevice() -- i.e. one that opens the specific (and well-known at that point) /dev/whatever and return the UnixFd. Then the whole method is a three-liner, and you can adjust it depending on the specific test case. This is the main idea how dbusmock, and avoids having to route every special case through upstream dbusmock releases :grin:

jadahl: Ah, neat. Do you have examples somewhere that illustrates how one does that? (add extra methods to an upstream template).

martinpitt: Sure! The [README](https://github.com/martinpitt/python-dbusmock#simple-example-from-shell) has an example in shell with `gdbus call`. [Here](https://github.com/martinpitt/python-dbusmock/blob/master/tests/test_api.py#L125) is an example for a multi-line function body in Python. But in the end it's really just calling `org.freedesktop.DBus.Mock.AddMethod(name, in_sig, out_sig, code)` on D-Bus. See the [AddMethod() documentation](https://github.com/martinpitt/python-dbusmock/blob/master/dbusmock/mockobject.py#L361)

jadahl: I would need to put the whole methods in a large string and send via AddMethod(). A bit annoying for "larger" functions like this but oh well. (unless there is some other way I haven't found yet).

martinpitt: If you really need a large method, then use a template. However, mocks are not *supposed* to have large methods -- possibly add some `assert()`s about the expected input, and return the mocked output value for the most part. In your case you have to open a /dev file and return the UnixFd, but that's just two or three lines then? This should not be too bad with a Python multi-line string?

jadahl: Yea, it's doable. The fact that I implemented it upstream first since I don't need to debug while writing a function inside a string. Thanks for the links!

jadahl: *the fact... helps, I mean

jadahl: Ended up like this, not so bad. Thanks for your help! For reference:

```diff
diff --git a/src/tests/meta-dbus-runner.py b/src/tests/meta-dbus-runner.py
index 69daa0af21..b16b0640b4 100755
--- a/src/tests/meta-dbus-runner.py
+++ b/src/tests/meta-dbus-runner.py
@@ -40,13 +40,12 @@ class MutterDBusTestCase(DBusTestCase):
         (klass.mocks_manager, klass.mock_obj) = klass.start_from_local_template(
             'meta-mocks-manager', {'templates-dir': get_templates_dir()})
 
-        logind = klass.start_from_template('logind')
         klass.start_from_local_template('localed')
 
         klass.system_bus_con = klass.get_dbus(system_bus=True)
         klass.session_bus_con = klass.get_dbus(system_bus=False)
 
-        klass.add_logind_session(logind)
+        klass.init_logind()
 
         if klass.session_bus_con.name_has_owner('org.gnome.Mutter.DisplayConfig'):
             raise Exception(
@@ -105,13 +104,50 @@ class MutterDBusTestCase(DBusTestCase):
         return mocks
 
     @classmethod
-    def add_logind_session(klass, logind):
+    def init_logind(klass):
+        logind = klass.start_from_template('logind')
+
         [p_mock, obj] = logind
+
         mock_iface = 'org.freedesktop.DBus.Mock'
         obj.AddSeat('seat0', dbus_interface=mock_iface)
-        obj.AddSession('dummy', 'seat0',
-            dbus.types.UInt32(os.getuid()), getpass.getuser(),
-            True, dbus_interface=mock_iface)
+        session_path = obj.AddSession('dummy', 'seat0',
+                                      dbus.types.UInt32(os.getuid()),
+                                      getpass.getuser(),
+                                      True,
+                                      dbus_interface=mock_iface)
+
+        session_obj = klass.system_bus_con.get_object('org.freedesktop.login1', session_path)
+        session_obj.AddMethod('org.freedesktop.login1.Session',
+                              'TakeDevice',
+                              'uu', 'hb',
+'''
+import re
+
+major = args[0]
+minor = args[1]
+
+sysfs_uevent_path = '/sys/dev/char/{}:{}/uevent'.format(major, minor)
+sysfs_uevent = open(sysfs_uevent_path, 'r')
+devname = None
+for line in sysfs_uevent.readlines():
+    match = re.match('DEVNAME=(.*)', line)
+    if match:
+        devname = match[1]
+        break
+sysfs_uevent.close()
+if not devname:
+    raise dbus.exceptions.DBusException(f'Device file {major}:{minor} doesn\\\'t exist',
+                                        major=major, minor=minor)
+fd = os.open('/dev/' + devname, os.O_RDWR | os.O_CLOEXEC)
+unix_fd = dbus.types.UnixFd(fd)
+os.close(fd)
+ret = (unix_fd, False)
+''')
+        session_obj.AddMethods('org.freedesktop.login1.Session', [
+            ('ReleaseDevice', 'uu', '', ''),
+            ('TakeControl', 'b', '', ''),
+        ])
 
     def wrap_call(self, args):
         env = {}
```

I'll close this PR, nothing here nees to go upstream I guess.

martinpitt: Nice! You could use `with open...` to avoid explicit .close() calls. You might also want to make your test stricter by asserting the values of major/minor, open the known VKMS device, and spare yourself the uevent parsing. But that's just minor optimization of course.

jadahl: Good tips! I'm afraid I have to deal with the uevent parsing though, e.g. I wouldn't here know whether I try to open the VKMS device or some evdev device created from uinput. Mode setting tests are passing, without this PR anyhow, so yay! \o/

PR #97, opened 2022-01-02
logind: additions and improvements for Lock/Unlock/LockedHint

andyholmes:
The two commits add the `LockedHint`/`SetLockedHint()` property/method and
amend the `Lock()`/`Unlock()` methods to emit their respective signals, as
described in the interface documentation.

### DIFF
--- a/dbusmock/templates/logind.py
+++ b/dbusmock/templates/logind.py
@@ -268,11 +268,11 @@ def AddSession(self, session_id, seat, uid, username, active):
                    [
                        ('Activate', '', '', ''),
                        ('Kill', 'ss', '', ''),
-                       ('Lock', '', '', ''),
+                       ('Lock', '', '', 'self.EmitSignal("", "Lock", "", [])'),
                        ('SetIdleHint', 'b', '', ''),
                        ('SetLockedHint', 'b', '', 'self.UpdateProperties("", {"LockedHint": args[0]})'),
                        ('Terminate', '', '', ''),
-                       ('Unlock', '', '', ''),
+                       ('Unlock', '', '', 'self.EmitSignal("", "Unlock", "", [])'),
                    ])
 
     # add session to seat
--- a/dbusmock/templates/logind.py
+++ b/dbusmock/templates/logind.py
@@ -241,6 +241,7 @@ def AddSession(self, session_id, seat, uid, username, active):
                        'ResetControllers': dbus.Array([], signature='s'),
                        'Active': active,
                        'IdleHint': False,
+                       'LockedHint': False,
                        'KillProcesses': False,
                        'Remote': False,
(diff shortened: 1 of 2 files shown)


dbusmock/templates/logind.py
```diff
@@ -269,6 +270,7 @@ def AddSession(self, session_id, seat, uid, username, active):
                        ('Kill', 'ss', '', ''),
                        ('Lock', '', '', ''),
                        ('SetIdleHint', 'b', '', ''),
+                       ('SetLockedHint', 'b', '', ''),
```
martinpitt, line 273: Shouldn't this actually set the `LockedHint` property? If you are testing this functionality, how does this make sense?

martinpitt (commented):
Thanks, and happy new year! :tada:

dbusmock/templates/logind.py
andyholmes: Happy new year to you, too! No wonder my test wasn't quite working; I guess that's my first face-palm of the year :)

martinpitt: Yes, that looks fine. Want to augment tests/test_logind.py's `test_session()` to check `loginctl show-session`'s output before/after calling SetLockedHint() on the mock? that would make sure that the properties are the ones that the real loginctl expects.

martinpitt (changes requested):
As said above, let's add some tests. I'm happy to do that for you if you don't feel comfortable with that, of course!

dbusmock/templates/logind.py
andyholmes: Okay, done :)

I also added a check for the `Lock()`/`Unlock()` methods, but as a separate commit in case those aren't necessary or correct.

tests/test_logind.py
```diff
@@ -118,6 +122,13 @@ def test_session(self):
         self.assertRegex(out, 'Id=c1')
         self.assertRegex(out, 'LockedHint=yes')
 
+        session_mock.Lock()
+        self.assertRegex(self.p_mock.stdout.read(),
+                         b'emit org.freedesktop.login1.Session.Lock\n')
```
martinpitt, line 127: This test isn't very interesting, as it does not check any behaviour through the official loginctl CLI; it just checks its own mock code. It doesn't hurt, of course, but that part is already well covered.

martinpitt (changes requested):
Thanks! Please squash the tests into their respective code change commits. I don't have a strong opinion about the Lock/Unlock signal test, you can either drop or squash it. After that, this should be good to go, all tests passed.

andyholmes:
Should be good to go, thanks for the review!

martinpitt (approved):
Perfect, thanks for this improvement and bearing with me!

PR #98, opened 2022-01-05
mockobject: Print the object path when manipulating properties

hadess:
This is invaluable when debugging property changes with multiple object
paths involved, such as bluez with multiple Bluetooth adapters.


martinpitt (approved):
en effet, merci !

PR #99, opened 2022-01-05
bluez: Fix stale Discoverable/Discovering property

hadess:
When starting a new discovery, we would send signals that the properties
were changed, but we never set those as properties internally, so anyone
fetching them would get the default value.

Modify the property values directly instead of using intermediate
variables.


martinpitt (approved):
Merci !

PR #100, opened 2022-01-17
bluez5: Add all missing device properties

hadess:
Export the same device properties as bluez would so they can be easily
updated by tests, eg. setting the Class or Icon.


hadess:
Needs a bit more work, the properties show up as the wrong type.

martinpitt (approved):
Merci !

PR #107, opened 2022-03-07
dbusmock: Do not register standard session service directories

benzea:
Otherwise GTK 4 might try to start the A11Y services, which may fail.

Same thing we already do for system services. Without this, I have the issue that GTK 4 tries to start the A11Y services. This works fine in the CI, but I suppose that is because the service is simply missing.

### DIFF
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -37,34 +37,19 @@ class DBusTestCase(unittest.TestCase):
     system_bus_pid = None
 
     @classmethod
-    def start_session_bus(cls) -> None:
+    def __start_bus(cls, bus_type) -> None:
         '''Set up a private local session bus
 
         This gets stopped automatically at class teardown.
         '''
-        (DBusTestCase.session_bus_pid, addr) = cls.start_dbus()
-        os.environ['DBUS_SESSION_BUS_ADDRESS'] = addr
-
-        cls.addClassCleanup(setattr, DBusTestCase, 'session_bus_pid', None)
-        cls.addClassCleanup(os.environ.pop, 'DBUS_SESSION_BUS_ADDRESS')
-        cls.addClassCleanup(cls.stop_dbus, DBusTestCase.session_bus_pid)
-
-    @classmethod
-    def start_system_bus(cls) -> None:
-        '''Set up a private local system bus
-
-        This gets stopped automatically at class teardown.
-        '''
-        # create a temporary configuration which makes the fake bus actually
-        # appear a type "system"
-        with tempfile.NamedTemporaryFile(prefix='dbusmock_cfg') as c:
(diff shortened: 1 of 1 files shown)


dbusmock/testcase.py
```diff
@@ -42,7 +42,24 @@ def start_session_bus(cls) -> None:
...
+
+  <policy context="default">
+    <allow send_destination="*" eavesdrop="true"/>
+    <allow eavesdrop="true"/>
+    <allow own="*"/>
+  </policy>
+</busconfig>
+''')
+            c.flush()
+            (DBusTestCase.session_bus_pid, addr) = cls.start_dbus(conf=c.name)
```
martinpitt, line 62: This whole part is now identical between start_{session,system}_bus(), aside from the comment. So let's factorize this. @benzea , I'm happy to do that if you don't have time.

martinpitt (changes requested):
Thanks! But let's clean this up a bit.

martinpitt:
pylint failure:

    dbusmock/testcase.py:40:4: W0238: Unused private member `DBusTestCase.__start_bus(cls, bus_type)` (unused-private-member)

dbusmock/testcase.py
```diff
@@ -40,16 +40,20 @@ class DBusTestCase(unittest.TestCase):
     def start_session_bus(cls) -> None:
         '''Set up a private local session bus
 
-        This gets stopped automatically in tearDownClass().
+        This gets stopped automatically at class teardown.
         '''
         (DBusTestCase.session_bus_pid, addr) = cls.start_dbus()
         os.environ['DBUS_SESSION_BUS_ADDRESS'] = addr
 
+        cls.addClassCleanup(setattr, DBusTestCase, 'session_bus_pid', None)
```
martinpitt: Nice! This is Python ≥ 3.8, last time I tried to introduce it it was still not available on all supported OSes. But seems CI is happy with it now (other than pylint).

benzea:
I just changed it to explicitly use `DBusTestCase.__start_bus`, kind of makes sense that way.

martinpitt (approved):
Cheers!

benzea:
Hrm, maybe this may have been a bad idea.

I did have the issue that in the g-c-c tests, GTK4 tried to start the A11Y services and that failed (causing a warning, causing a crash).

But, now my g-s-d tests are failing because dconf is not running. That is easily fixable though (just explicitly launch `/usr/libexec/dconf-service`).

So, not quite sure what is the right thing for now. Maybe we should have our own service directory and copy some common stuff over?

martinpitt:
@benzea : My gut feeling is that it's right to not auto-launch services for unit tests. It feels like this should happen explicitly, so that tests control what is running and nothing unexpected happens. Creating the test's own service directory and adding a convenience API to install symlinks into it sounds like a good idea, though!

PR #108, opened 2022-03-15
dbusmock: Allow creating/enabling .service files

benzea:
We now provide a classmethod to get the service directory of the dbus
server that will be launched. Also new is the enable_service function
which will symlink the service in order to enable it in the test
environment.

---

This is working great for the g-s-d tests. The attribute is a bit ugly, but `cls` can be either `DBusTestCase` or any subclass, making it unpredictable if it is first called with a derived class as an argument. As such, the code just hardcodes `DBusTestCase` (and manually mangles the name for the delattr).

### DIFF
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -11,6 +11,7 @@ __copyright__ = '(c) 2012 Canonical Ltd.'
 
 import errno
 import os
+import shutil
 import signal
 import subprocess
 import sys
@@ -35,6 +36,29 @@ class DBusTestCase(unittest.TestCase):
     '''
     session_bus_pid = None
     system_bus_pid = None
+    _DBusTestCase__datadir = ''
+
+    @classmethod
+    def get_services_dir(cls, system_bus: bool = False) -> str:
+        '''Returns the private services directory for the bus type in question.
+        This allows dropping in a .service file so that the dbus server inside
+        dbusmock can launch it.
+        '''
+        # NOTE: Explicitly use the attribute of DBusTestCase, as cls may be a
+        # different class depending on how the method is called.
+        if system_bus:
+            services_dir = 'system_services'
+        else:
+            services_dir = 'services'
+        if not DBusTestCase._DBusTestCase__datadir:
+            DBusTestCase._DBusTestCase__datadir = tempfile.mkdtemp(prefix='dbusmock_data_')
+            cls.addClassCleanup(setattr, DBusTestCase, '_DBusTestCase__datadir', '')
+            cls.addClassCleanup(shutil.rmtree, DBusTestCase._DBusTestCase__datadir)
+
+            os.mkdir(os.path.join(DBusTestCase._DBusTestCase__datadir, 'system_services'))
+            os.mkdir(os.path.join(DBusTestCase._DBusTestCase__datadir, 'services'))
+
+        return os.path.join(DBusTestCase._DBusTestCase__datadir, services_dir)
 
     @classmethod
     def __start_bus(cls, bus_type) -> None:
@@ -42,14 +66,17 @@ class DBusTestCase(unittest.TestCase):
 
         This gets stopped automatically at class teardown.
         '''
-        with tempfile.NamedTemporaryFile(prefix=f'dbusmock_{bus_type}_cfg', mode='w') as c:
+        cls.get_services_dir()
+
+        with open(os.path.join(DBusTestCase._DBusTestCase__datadir, f'dbusmock_{bus_type}_cfg'), 'w', encoding='ascii') as c:
             c.write(f'''<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-Bus Bus Configuration 1.0//EN"
  "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
 <busconfig>
   <type>{bus_type}</type>
   <keep_umask/>
   <listen>unix:tmpdir=/tmp</listen>
-  <!-- We do not add standard_{bus_type}_servicedirs (i.e. we have *no* service directory). -->
+  <!-- We do not add standard_{bus_type}_servicedirs (i.e. we only have our private services directory). -->
+  <servicedir>{cls.get_services_dir(bus_type == 'system')}</servicedir>
 
   <policy context="default">
     <allow send_destination="*" eavesdrop="true"/>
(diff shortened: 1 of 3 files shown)


benzea:
Note that the accountsservice tests seem to be completely broken. Maybe they should just be removed?

dbusmock/testcase.py
```diff
@@ -251,3 +278,30 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+
+        Returns True if a service file was found and symlinked.
+        '''
+        if system_bus:
+            services_dir = 'system-services'
+        else:
+            services_dir = 'services'
+        xdg_data_dirs = '/usr/local/share/:/usr/share/'
+        if 'XDG_DATA_DIRS' in os.environ and os.environ['XDG_DATA_DIRS']:
+            xdg_data_dirs = os.environ['XDG_DATA_DIRS']
```
martinpitt, line 300: Please write this as:

   xdg_data_dirs = os.environ.get('XDG_DATA_DIRS', '/usr/local/share/:/usr/share/')

martinpitt (changes requested):
Thanks! This looks fine to me, but it's non-trivial logic and I'd like to have it tested. Want to write  test for it (add e.g. a d-conf dependency), or shall I look into that?

It would also be good to add a stanza about the private bus and services dirs to README (but I can do that as a follow-up, too)

dbusmock/testcase.py
benzea: Unfortunately, that wouldn't conform to the XDG specification. i.e. an empty XDG_DATA_DIRS also needs to be handled as the default value.

martinpitt: OK, then

    os.environ.get('XDG_DATA_DIRS') or '/usr/local/share/:/usr/share/'

?

benzea:
Sorry, another attempt at this. I realised that being able to disable a service is kinda sensible in the test. And followed up by adding the appropriate SIGHUP to reload the configuration.

dbusmock/testcase.py
```diff
@@ -251,3 +278,58 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+        for d in xdg_data_dirs.split(':'):
+            src = os.path.join(d, 'dbus-1', services_dir, service + '.service')
+            if os.path.exists(src):
+                os.symlink(src, os.path.join(cls.get_services_dir(system_bus), service + '.service'))
+                # Poke dbus to reload the services
+                if dbus_pid:
+                    os.kill(dbus_pid, signal.SIGHUP)
+                return True
+
+        return False
```
martinpitt, line 311: Feels more pythonic and it's also more robust to raise an exception if the service wasn't found -- that's normally unexpected in a unit test, so it shouldn't require additional code to handle that condition, but fail instead.

(You can do this very elegantly with an `else:` on the `for` loop)

dbusmock/testcase.py
```diff
@@ -251,3 +278,58 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+            services_dir = 'services'
+            dbus_pid = cls.session_bus_pid
+        xdg_data_dirs = os.environ.get('XDG_DATA_DIRS') or '/usr/local/share/:/usr/share/'
+
+        for d in xdg_data_dirs.split(':'):
+            src = os.path.join(d, 'dbus-1', services_dir, service + '.service')
+            if os.path.exists(src):
+                os.symlink(src, os.path.join(cls.get_services_dir(system_bus), service + '.service'))
+                # Poke dbus to reload the services
+                if dbus_pid:
```
martinpitt, line 307: Likewise, this function should crash if there's no bus pid, i.e. `assert` this.

dbusmock/testcase.py
```diff
@@ -251,3 +278,58 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+        else:
+            dbus_pid = cls.session_bus_pid
+
+        try:
+            os.unlink(os.path.join(cls.get_services_dir(system_bus), service + '.service'))
+        except OSError:
+            return False
+
+        if dbus_pid:
+            os.kill(dbus_pid, signal.SIGHUP)
```
martinpitt, line 333: Likewise, please drop the try, if, and return value.

tests/test_api.py
```diff
@@ -912,6 +913,82 @@ def __init__(self):
...
+        session_dir = os.path.join(cls.xdg_data_dir, 'dbus-1', 'services')
+        os.mkdir(system_dir)
+        os.mkdir(session_dir)
+
+        with open(os.path.join(system_dir, 'org.TestSystem.service'), 'w', encoding='ascii') as s:
+            s.write('[D-BUS Service]\n' +
+                    'Name=org.TestSystem\n'
+                    'Exec=/usr/bin/python3 -c "import sys; from gi.repository import GLib, Gio; '
+                    '     Gio.bus_own_name(Gio.BusType.SYSTEM, \'org.TestSystem\', 0, None, None, lambda *args: sys.exit(0)); '
+                    '     GLib.MainLoop().run()"\n'
```
martinpitt, line 937: Hah, good one! It's a pleasure how compact it is in Python.

Please use a `"""` multiline string for this, which avoids the (inconsistent) `+`, `\n`, and quoted single quote.

tests/test_api.py
```diff
@@ -912,6 +913,82 @@ def __init__(self):
...
+        cls.start_system_bus()
+        cls.start_session_bus()
+
+    def test_session_service_isolation(self):
+        dbus_con = self.get_dbus(system_bus=False)
+        dbus_obj = dbus_con.get_object('org.freedesktop.DBus', '/org/freedesktop/DBus')
+        dbus_if = dbus.Interface(dbus_obj, 'org.freedesktop.DBus')
+
+        self.assertEqual(dbus_if.ListActivatableNames(), ['org.freedesktop.DBus'])
+        self.enable_service('org.TestSession')
```
martinpitt, line 957: E.g. this doesn't even assert the return value (and really shoudn't).

martinpitt (changes requested):
Nice, thanks! Some small cleanups, then this is ready indeed. After that I think we should do a new release?

dbusmock/testcase.py
benzea: Ah, this one is fine, I think. i.e. it is legal to enable a service before starting dbus.

tests/test_api.py
benzea: ah, I tried. But the script line is way too long and then the code tests complain about that.

I could add a comment to ignore the warning instead.

benzea:
Really? 

```
tests/test_api.py:1:0: C0302: Too many lines in module (1001/1000) (too-many-lines)
[741](https://github.com/martinpitt/python-dbusmock/runs/5591933450?check_suite_focus=true#step:3:741)
```

martinpitt:
@benzea Err, yes, that's a bit stupid. Let's just disable that error in tests/test_code.py:48.

However, [this failure](https://github.com/martinpitt/python-dbusmock/runs/5591933315?check_suite_focus=true) is different, suggests a race condition?

benzea:
> @benzea Err, yes, that's a bit stupid. Let's just disable that error in tests/test_code.py:48.
> 
> However, [this failure](https://github.com/martinpitt/python-dbusmock/runs/5591933315?check_suite_focus=true) is different, suggests a race condition?

Hrm, yeah, that is a proper race condition. I suppose we should ping the dbus server once after sending SIGHUP.

dbusmock/testcase.py
```diff
@@ -251,3 +278,56 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+    def enable_service(cls, service, system_bus: bool = False) -> None:
+        '''Enable the given well known service name inside dbusmock
+
+        This symlinks a service file from the usual dbus service directories
+        into the dbusmock environment. Doing that allows the service to be
+        launched automatically if they are defined within $XDG_DATA_DIRS.
+
+        The daemon configuration is reloaded if a test bus is running.
+
+        Returns True if a service file was found and symlinked.
```
martinpitt, line 292: This line is obsolete.

dbusmock/testcase.py
```diff
@@ -251,3 +278,56 @@ def spawn_server_template(cls, template: str, parameters: Dict[str, Any] = None,
...
+            dbus_if.ReloadConfig()
+
+    @classmethod
+    def disable_service(cls, service, system_bus: bool = False) -> None:
+        '''Disable the given well known service name inside dbusmock
+
+        This unlink's the .service file for the service and reloads the
+        daemon configuration if a test bus is running.
+
+        Returns True if a service was disabled
```
martinpitt, line 320: Dito

martinpitt (approved):
Thanks! This looks really good, just the documentation needs a bit of fixing. For efficiency I'll just force-push that onto your branch (as it's super-obvious), and land this. I'll do a follow-up with some README.

martinpitt (approved):

PR #109, opened 2022-03-18
Ensure method calls from templates are logged

whot:
For interfaces loaded via a template, calls to static methods were never
logged and the MethodCalled signal was never emitted. Fix this by
wrapping the function from the module in what is a "standard" Python
function debug decorator, except that it adds the calls to the log
instead.

This is a bit awkward since the first argument is always self - we need
to pass that on to the function but we don't want to log it.

*Note: not 100% sure that's the right approach but couldn't think of anything else*

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -11,6 +11,7 @@ __author__ = 'Martin Pitt'
 __copyright__ = '(c) 2012 Canonical Ltd.'
 
 import copy
+import functools
 import importlib
 import importlib.util
 import os
@@ -118,6 +119,22 @@ def _wrap_in_dbus_variant(value):
     if isinstance(value, str):
         return dbus.String(value, variant_level=1)
     raise dbus.exceptions.DBusException(f'could not wrap type {type(value)}')
+
+
+def loggedmethod(self, func):
+    """Decorator for a method to end in the call log"""
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        fname = func.__name__
+        log_args = args[1:]  # args[0] is self
+        self.log(fname + _format_args(log_args))
+        self.call_log.append((int(time.time()), fname, log_args))
+        self.MethodCalled(fname, log_args)
+
+        return func(*args, **kwargs)
+
+    return wrapper
 
 
 class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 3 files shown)


dbusmock/mockobject.py
```diff
@@ -548,6 +565,9 @@ def AddTemplate(self, template: str, parameters: PropsType) -> None:
             # pylint: disable=protected-access
             fn = getattr(module, symbol)
             if ('_dbus_interface' in dir(fn) and ('_dbus_is_signal' not in dir(fn) or not fn._dbus_is_signal)):
+
```
martinpitt, line 568: Let's drop this excess empty line.

martinpitt (changes requested):
Many thanks @whot! I like this in general. It is a bit hackish indeed, but mockobject.py is already rummaging around in dbus-python's guts a lot anyway.

However, it [breaks the ofono tests](https://github.com/martinpitt/python-dbusmock/runs/5623295144?check_suite_focus=true#step:3:1358):
```
  File "/tmp/source/tests/test_ofono.py", line 34, in setUpClass
    (cls.p_mock, cls.obj_ofono) = cls.spawn_server_template(
  File "/tmp/source/dbusmock/testcase.py", line 250, in spawn_server_template
    obj.AddTemplate(template, parameters,
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/service.py", line 715, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/tmp/source/dbusmock/mockobject.py", line 581, in AddTemplate
    module.load(self, parameters)
  File "/tmp/source/dbusmock/templates/ofono.py", line 48, in load
    mock.AddModem(parameters.get('ModemName', 'ril_0'), {})
  File "/tmp/source/dbusmock/mockobject.py", line 133, in wrapper
    self.MethodCalled(fname, log_args)
  File "/usr/lib/python3/dist-packages/dbus/decorators.py", line 325, in emit_signal
    message.append(signature=signature, *args)
ValueError: Unable to guess signature from an empty dict
```

I'll have a look at this.

dbusmock/mockobject.py
martinpitt: Done.

martinpitt:
I added a second commit which fixes/works around the problem. It is sort of an API break that automatic signature introspection does not work any more now with static methods, but small enough to be acceptable IMHO (there won't be that many custom templates out there with that problem).

martinpitt (approved):
Thanks!

whot:
thanks, much appreciated!

heftig:
This seems to be pretty broken, see #118.

PR #117, opened 2022-03-22
Small mock object improvements

benzea:
Had some trouble with the g-s-d tests, here some patches that are hopefully useful.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -683,36 +683,40 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         This gets "instantiated" in AddMethod(). Execute the code snippet of
         the method and return the "ret" variable if it was set.
         '''
-        # print('mock_method', dbus_method, self, in_signature, args, kwargs, file=sys.stderr)
+        # print('mock_method', dbus_method, self, in_signature, args, _, file=sys.stderr)
 
         try:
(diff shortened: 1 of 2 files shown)


dbusmock/mockobject.py
```diff
@@ -686,9 +686,12 @@ def mock_method(self, interface: str, dbus_method: str, in_signature: str, *args
...
             raise TypeError('Fewer items found in D-Bus signature than in Python arguments')
-        m = dbus.connection.MethodCallMessage('a.b', '/a', 'a.b', 'a')
-        m.append(signature=in_signature, *args)
-        args = m.get_args_list()
+        try:
+            m = dbus.connection.MethodCallMessage('a.b', '/a', 'a.b', 'a')
+            m.append(signature=in_signature, *args)
+            args = m.get_args_list()
+        except TypeError as e:
+            raise dbus.exceptions.DBusException('Invalid arguments', name='org.freedesktop.DBus.Error.InvalidArgs')
```
martinpitt: Makes sense. Can you please cover this in test_api.py?

martinpitt (changes requested):
Nice improvements, thanks! Please add unit tests, and there is one pylint and one mypy error. I'm happy to help with this if you don't have time.

dbusmock/mockobject.py
benzea: Updated the test, also changed the signal emitting to match now.

PR #119, opened 2022-04-04
Ignore a missing self.logfile on errors

whot:
If the `super()` constructor `dbus.service.Object.__init__` raises an
exception, we end up with `self.logfile` never being assigned. This causes
an AttributeError during `__del__`, obscuring the actual issues in the
code.


martinpitt (approved):
Thanks!

Issue #123, opened 2022-04-04
0.27.4 causes gnome-bluetooth 42 tests to fail


smcv:
gnome-bluetooth version 42 (Debian source package gnome-bluetooth3) passes its unit tests with python3-dbusmock 0.27.3-1 but fails with 0.27.4-1. This appears to have been caused by the more thorough D-Bus signature checking introduced to fix #118.

```
17/17 Tests.test_pairing                 FAIL            5.91s   exit status 1
17:32:44 top_srcdir=/<<PKGBUILDDIR>> top_builddir=/<<PKGBUILDDIR>>/obj-x86_64-linux-gnu MALLOC_PERTURB_=88 /usr/bin/python3 /<<PKGBUILDDIR>>/obj-x86_64-linux-gnu/../tests/integration-test.py Tests.test_pairing
----------------------------------- output -----------------------------------
stdout:
1649093564.622 AddAdapter "hci0" "my-computer"
1649093564.623 emit / org.freedesktop.DBus.ObjectManager.InterfacesAdded "/org/bluez/hci0" {"org.bluez.Adapter1": {"UUIDs": ["00001200-0000-1000-8000-00805f9b34fb", "00001800-0000-1000-8000-00805f9b34fb", "00001801-0000-1000-8000-00805f9b34fb", "0000110e-0000-1000-8000-00805f9b34fb", "0000110c-0000-1000-8000-00805f9b34fb"], "Discoverable": False, "Discovering": False, "Pairable": True, "Powered": True, "Address": "00:01:02:03:04:05", "AddressType": "public", "Alias": "my-computer", "Modalias": "usb:v1D6Bp0245d050A", "Name": "my-computer", "Class": 268, "DiscoverableTimeout": 180, "PairableTimeout": 0}}
1649093564.624 AddDevice "hci0" "11:22:33:44:55:66" "My Phone"
[...]

Issue #118, opened 2022-04-02
0.27.3 breaks most accountsservice tests


heftig:
I've updated from 0.25.0 to 0.27.3. Since then, the accountsservice tests are broken.

E.g.:

```
10/12 accountsservice:accounts-service+libaccountsservice+TestAccountsService / TestAccountsService.test_recreate_user            FAIL            0.32s   exit status 1
>>> GI_TYPELIB_PATH=/build/accountsservice/src/build/src/libaccountsservice G_DEBUG=fatal-criticals G_MESSAGES_DEBUG=all MALLOC_PERTURB_=56 LD_LIBRARY_PATH=/build/accountsservice/src/build/src/libaccountsservice /usr/bin/python3 /build/accountsservice/src/build/../accountsservice/tests/test-libaccountsservice.py TestAccountsService.test_recreate_user
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
test_recreate_user (__main__.TestAccountsService) ... (process:2193): accountsservice-DEBUG: 16:08:09.032: Finding a graphical session for user 1042
(process:2193): accountsservice-DEBUG: 16:08:09.032: Failed to identify the current session
(process:2193): accountsservice-DEBUG: 16:08:09.032: ActUserManager: seat unloaded, so trying to set loaded property
(process:2193): accountsservice-DEBUG: 16:08:09.032: ActUserManager: Seat wouldn't load, so giving up on it and setting loaded property
(process:2193): accountsservice-DEBUG: 16:08:09.032: ActUserManager: already loaded, so not setting loaded property
(process:2193): accountsservice-DEBUG: 16:08:09.032: ActUserManager: Creating user 'pizza', 'I Love Pizza', 1
(process:2193): accountsservice-DEBUG: 16:08:09.033: ActUserManager: calling 'ListCachedUsers'
(process:2193): accountsservice-DEBUG: 16:08:09.034: ActUserManager: ListCachedUsers finished with empty list, maybe setting loaded property now
(process:2193): accountsservice-DEBUG: 16:08:09.034: ActUserManager: already loaded, so not setting loaded property
ERROR

======================================================================
ERROR: test_recreate_user (__main__.TestAccountsService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/accountsservice/src/build/../accountsservice/tests/test-libaccountsservice.py", line 111, in test_recreate_user
[...]

PR #124, opened 2022-04-04
dbusmock.templates.bluez5: Don't rely on a default value for an argument

smcv:
With the more thorough signature checking from #118, the gnome-bluetooth
unit tests fail, because Pair() called PairDevice() with 2 arguments
but PairDevice() expects 3 arguments. Optional arguments with a default
are not really something that exists in the interoperable subset of
D-Bus.

Resolves: #123


martinpitt (approved):
Thanks, and sorry for the trouble! As I mentinoed in #123, there is some more fallout from this. I'll land this fix now, enable/fix the bluez pairing tests, and also look at the corresponding issue in accountsservice.

PR #126, opened 2022-05-19
testcase: Throw an error when spawning a well-known name that exists

benzea:
It is expected that the name is not yet owned. And if we continue, we'll
likely try to re-initialize an existing object that should have been
cleaned up. If this happens, the likely explanation is that a previous
test left the mock environment in a bad state, so error out immediately
with an error that is easier to understand.


martinpitt (approved):
Danke!

PR #128, opened 2022-05-27
tests: remove the accountsservice tests

### LINKED ISSUE #75: AccountServices mock test hangs

whot:
These tests hang, no-one's been able to find out why, there are some
oddities going on anyway and meanwhile this is blocking people from
running the test suite.

These tests also appear to be largely identical to the ones in the
accountsservice repo, so let's drop them here.
https://gitlab.freedesktop.org/accountsservice/accountsservice/-/blob/main/tests/test-libaccountsservice.py

Fixes #75


martinpitt:
@3v1n0 : Are you okay with dropping the accountsservice template from python-dbusmock? Indeed this has been bitrotting for a while, and I haven't been able to fix this -- this needs a domain expert. Just dropping the tests doesn't help anyone, as the template itself is currently broken beyond repair.

@whot : Can you please drop the corresponding template as well? I'm also happy to do this by force-pushing to your branch.

whot:
> @whot : Can you please drop the corresponding template as well? I'm also happy to do this by force-pushing to your branch.

done!

martinpitt (approved):
Thanks! I'll adjust the commit message on merging accordingly.

PR #129, opened 2022-05-27
Allow setting the destination on signals

whot:
Sits on top of ~~#128~~ #130 

~~Requires dbus-python to update first: https://gitlab.freedesktop.org/dbus/dbus-python/-/merge_requests/13~~

Switch to use dbus-python's `SignalMessage` directly for our signal emission instead of wrapping the `@signal` decorator.

Add a new DBus method EmitSignalDetailed that also takes an a{sv} of
magic keywords for details to be passed down.

The only supported detail right now is "destination" for the destination ~~, using
the new destination_keyword from dbus-python, see
https://gitlab.freedesktop.org/dbus/dbus-python/-/merge_requests/18~~

This allows us to send signals to one specific target even if they don't
have a match rule setup (used by libportal for example).

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -607,6 +607,21 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         self._template = template
         self._template_parameters = parameters
 
+    def _emit_signal(self, interface: str, name: str, signature: str, sigargs: Tuple[Any, ...]) -> None:
+        # pylint: disable=protected-access
+        if not interface:
+            interface = self.interface
+
+        args = _convert_args(signature, sigargs)
+
+        sig = dbus.lowlevel.SignalMessage(self.path, interface, name)
+        sig.append(*args, signature=signature)
+
+        for location in self.locations:
+            conn = location[0]
+            conn.send_message(sig)
+        self.log(f'emit {self.path} {interface}.{name}{_format_args(args)}')
+
     @dbus.service.method(MOCK_IFACE,
                          in_signature='sssav',
                          out_signature='')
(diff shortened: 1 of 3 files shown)


3v1n0:
I wanted to do the same, but via [dbus-python support first](https://gitlab.freedesktop.org/dbus/dbus-python/-/merge_requests/13), so maybe can join the efforts to get that upstreamed first?

whot:
New attempt that decouples us from dbus-python's requirements: use the `dbus.lowlevel.SignalMessage` directly instead of wrapping the `@signal` function decorator because that allows us access to the signal directly, so we can set the destination without dbus-python changes.

The signal code is effectively the same as dbus-python's code for the decorator. Would this work?

3v1n0:
I'd say yes. And seen the python-dbus upstream state I think it's probably the best route to go, also to make it work without requiring an updated library.

dbusmock/mockobject.py
```diff
@@ -607,6 +607,21 @@ def AddTemplate(self, template: str, parameters: PropsType) -> None:
...
+        # pylint: disable=protected-access
+        if not interface:
+            interface = self.interface
+
+        args = _convert_args(signature, sigargs)
+
+        sig = dbus.lowlevel.SignalMessage(self.path, interface, name)
+        sig.append(*args, signature=signature)
+
+        for location in self.locations:
```
martinpitt, line 623: Note to self: This is https://dbus.freedesktop.org/doc/dbus-python/dbus.service.html#dbus.service.FallbackObject.locations

martinpitt (approved):
Very cool, thanks @whot for figuring this out! Big +1, this is much more flexible, and not even more complicated.

PR #130, opened 2022-06-08
tests: disable the "too many statements" error in pylint

whot:
For tests, we can easily run past the limit of 50 and it doesn't
necessarily indicate bad code quality.


martinpitt (approved):
Agreed, this check is rather dubiuos, similarly to "too many public methods". The "R8081" was for older pylints, but apparently that's gone now, so we don't need a numerical counterpart for "too many statements".

PR #136, opened 2022-06-28
dbusmock: drop __main__ dependency on gi.repository

allisonkarlitskaya:
Instead of iterating `GMainContext` via `gi.repository`, just use `ctypes` for the one function we need.
    
This makes it a lot easier to run dbusmock tests in a virtualenv.


dbusmock/__main__.py
```diff
@@ -117,4 +117,7 @@ def parse_args():
         main_object.AddTemplate(args.template, parameters)
 
     dbusmock.mockobject.objects[args.path] = main_object
-    main_loop.run()
+
+    context = GLib.MainContext.default()
+    while should_run:
```
martinpitt, line 122: `should_run` is never changed; that's normally done through g_main_loop_quit(), except that nothing in our mocks calls that (a mock is simply terminated). So why not a simple `while True:` here?

allisonkarlitskaya: `should_run` is absolutely changed via `should_run.pop` as the signal receiver for `Disconnected`.

martinpitt: Argh, of course, sorry.

martinpitt:
So this leaves the GLib usage in the logind and systemd templates and the tests. But I suppose these are not blocking you?

I'm happy to land that after a rebase. Thanks!

martinpitt (approved):
Thank you!

PR #140, opened 2022-07-17
Fix missing __version__ attribute

benjamb:
Without this autotools based version checking fails.

More specifically, checking the version with AX_PYTHON_MODULE_VERSION and version 0.28.2 installed will fail as it expects the '__version__' attribute to exist (this was removed in a6130f0490e95783619cd75d37e2e1e12364d51a when adding setuptools-scm): https://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_python_module_version.m4

This patch tells setuptools_scm to write the version out to a file, then imports the version from that file.

### DIFF
--- a/dbusmock/__init__.py
+++ b/dbusmock/__init__.py
@@ -7,8 +7,11 @@
 # later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
 # of the license.
 
+from .version import version
+
 __author__ = 'Martin Pitt'
 __copyright__ = '(c) 2012 Canonical Ltd.'
+__version__ = version
 
 from dbusmock.mockobject import (DBusMockObject, MOCK_IFACE,
                                  OBJECT_MANAGER_IFACE, get_object, get_objects)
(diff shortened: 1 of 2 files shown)


martinpitt:
Thanks @benjamb ! I just noticed the same after uploading the new version to [Debian](https://tracker.debian.org/pkg/python-dbusmock) -- it also [breaks ](https://ci.debian.net/data/autopkgtest/testing/amd64/u/upower/23771197/log.gz) upower's tests as it also queries `dbusmock.__version__`. The unit tests fail because the new `version.py` is apparently not shipped in tarballs, this still needs to be fixed. I'll push into your branch if you don't mind.

martinpitt:
Meh, this is quite messy.. I had to bring back setup.py, because pyproject.toml is still ridiculously broken in all distros, including Fedora rawhide. The idea is great, but it still needs a year or so to mature before it actually works :cry: So neither RPM builds nor running tests from git currently works. So TBH I'd rather go back to just statically putting the version into dbusmock/__init__.py for the time being, and read it in setup.py.

martinpitt:
I sent #141 with a more boring, but hopefully working approach, but linked to this so that I can give this another go in a while. Thank you!

benjamb:
@martinpitt Ouch, sounds like quite the nightmare! Thanks for fixing.

PR #141, opened 2022-07-17
Bring back dbusmock.__version__, revert sdist build to setup.py

martinpitt:
autotools's version checks or upower's test suite rely on this, so this
is API. Add a unit test to ensure it exists.

This should ideally be generated automatically [1], but that does not
work as long as packit, rpm builds etc. don't get along with pyproject.

So for the time being, move the static version back to
dbusmock/__init__.py, and read it from setup.py instead.

[1] https://github.com/martinpitt/python-dbusmock/pull/140

----

See [0.28.2 Debian regression](https://tracker.debian.org/pkg/python-dbusmock), breaks upower. Also thanks to @benjamb for his attempt in #140!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
       - '[0-9]*'
 jobs:
   release:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


martinpitt:
Meh, this failed again in Debian unstable/ubuntu devel. Not only in this PR, but already [yesterday](https://github.com/martinpitt/python-dbusmock/actions/runs/2683044249). I'll revert that bit as well -- after all, the official releases also use `setup.py sdist`.

PR #144, opened 2022-07-19
setup.py: Fix ModuleNotFoundError

### LINKED ISSUE #142: setup.py does not set sys.path correctly for importing dbusmock (when using PEP517 backend)
When using PEP517 backend, there is no guarantee that `.` will be present on `sys.path`. This causes e.g. the following build failure (or even worse, silently using version from installed dbusmock):

```pytb
$ gpep517 build-wheel --output-fd 1 --wheel-dir /tmp
  File "/usr/lib/python-exec/python3.11/gpep517", line 4, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/gpep517/__main__.py", line 191, in main
    return func(args)
           ^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/gpep517/__main__.py", line 87, in build_wheel
    wheel_name = backend.build_wheel(args.wheel_dir, args.config_json)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/setuptools/build_meta.py", line 244, in build_wheel
    return self._build_with_temp_dir(['bdist_wheel'], '.whl',
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/setuptools/build_meta.py", line 229, in _build_with_temp_dir
    self.run_setup()
  File "/usr/lib/python3.11/site-packages/setuptools/build_meta.py", line 174, in run_setup
    exec(code, locals())
  File "<string>", line 9, in <module>
ModuleNotFoundError: No module named 'dbusmock'
```

Will submit a PR shortly.

martinpitt:
Commit 842d2506fb71 broke `pip install` and also some variants of
PEP-517 build tools: `.` is not in their default `sys.path` any more.
Merely adding that to the search path is not sufficient: building an
sdist in a venv is tripping over the missing `import dbus`, as
dbus-python is just a runtime dependency, not a build-time one.

So instead of importing `dbusmock.__version__` in setup.py, ask
setuptools to dynamically generate the version in a declarative way, in
setup.cfg.

Many thanks to Michał Górny for analyzing this and proposing this
solution!

Fixes #142

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[metadata]
+version = attr: dbusmock.__version__
--- a/setup.py
+++ b/setup.py
@@ -6,14 +6,11 @@
 
 import setuptools
 
-from dbusmock import __version__
-
 with open('README.md', encoding="UTF-8") as f:
     readme = f.read()
 
 setuptools.setup(
     name='python-dbusmock',
-    version=__version__,
     description='Mock D-Bus objects',
     long_description=readme,
     long_description_content_type='text/markdown',


martinpitt:
I tested this with both `python3 -m build --sdist` and `python3 setup.py sdist`. @mgorny , WDYT?

mgorny (approved):
LGTM. BTW, you may consider switching fully to `setup.cfg`:
https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

It's there since 2017, so I think it should be safe to use by now ;-). It can also replace reading `README.md` manually.

martinpitt:
@mgorny: Ack, thanks! I'll look into that. I also want to add a test for `python3 -m build --sdist` to make sure that this keeps working from now on. I'll do that in a new PR.

PR #153, opened 2022-07-19
Drop setup.{cfg,py} and RHEL 9 support

martinpitt:
Draft for now, I want to keep this open to track the packaging/distro bugs which prevent this.

With a simple `packit srpm` and `rpmbuild -rb python-dbusmock-0.28*.src.rpm` in current Fedora 36 I get a broken rpm which only contains

     /usr/lib/python3.10/site-packages/UNKNOWN-0.0.post1+gac9a8ba.dist-info

and no actual modules. It seems it does not read the metadata from pyproject.toml properly.


martinpitt:
At least in a local rpm mock build, the version is still broken. The included dbusmock/_version.py seems correct, though.

martinpitt:
Better, but still blocked on the too old version on centos 9 stream, which misses a line break. I'll try to remove the docstring and pacify it in pylint instead.

martinpitt:
Dang, the [c9s build](https://artifacts.dev.testing-farm.io/f2bf60c5-ba8f-4458-9b4c-34f0622a0892/) is still an empty rpm; reproducible locally with 
```
packit srpm && mock -r centos-stream-9-x86_64 --no-bootstrap-image --isolation=simple *.src.rpm
less /var/lib/mock/centos-stream-9-x86_64/result/python3-dbusmock*.noarch.rpm
```

martinpitt:
I broke out the preparation commits into #219, they are fine.

martinpitt:
Let's see if I can add a RHEL 9 hack to the spec file so that the rpm won't be (almost) empty. I think this is just a spec problem, hopefully not a pybuild one.

martinpitt:
Eh, packit, where are you?

PR #157, opened 2022-08-06
Use setuptools_scm to determine version

martinpitt:
Determine dbusmock.__version__ from package metadata instead of
hardcoding it. With that, the release process is simply "push a signed
tag".

Drop MANIFEST.in. setuptools-scm automatically includes all git-tracked
files into the tarball.

See https://github.com/pypa/setuptools_scm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /dist
+/dbusmock/_version.py
 __pycache__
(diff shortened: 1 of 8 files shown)


martinpitt:
Closing, packit got confused over the master -> main rename.

PR #159, opened 2022-08-23
mockobject: allow sending signals from a specific path

whot:
While we can easily `mock.AddObject()` an object with a different path,
sending a signal from that object is more complicated. And even there it
doesn't cover the case where we want to fake a signal from an object
created by someone else.

So, let's fix this by adding a path argument to EmitSignalDetailed.

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -616,7 +616,8 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
 
         args = _convert_args(signature, sigargs)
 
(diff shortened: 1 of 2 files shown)


whot:
ftr, my use-case here is xdg-desktop-portal where the mock has to create Request and Session sub-objects. Being able to supply the signal path makes life a bit easier.

martinpitt (approved):
Whoops, sorry that this slipped through the cracks! (This was during my summer holidays when I unsubscribed from GitHub). This looks great, thank you!

PR #166, opened 2022-10-25
Fix OBEX PullAll after pathlib conversion

heftig:
The OBEX tests in folks fail with:

    (/build/folks/src/build/tests/bluez/bluez-individual-retrieval:3488):
    folks-WARNING **: 17:30:47.981:
    Error preparing persona store ‘bluez:00:00:00:00:00:00’:
    The OBEX address book transfer from device ‘My Phone’ failed:
    GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs:
    Invalid arguments: Expected a string or unicode object

`PullAll` passes a `pathlib.Path` to `manager.EmitSignal`. Converting it to a `str` fixes the issue.

### DIFF
--- a/dbusmock/templates/bluez5-obex.py
+++ b/dbusmock/templates/bluez5-obex.py
@@ -225,7 +225,7 @@ def PullAll(self, target_file, filters):
 
     # Emit a behind-the-scenes signal that a new transfer has been created.
     manager.EmitSignal(OBEX_MOCK_IFACE, 'TransferCreated', 'sa{sv}s',
-                       [transfer_path, filters, filename])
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Oops, vielen Dank! That's indeed the one bluez test that CI skips, as pbap-client is not packaged (and thus hard to get).

martinpitt:
@heftig : How urgently do you need this in distros and/or pip? Or are you checking it out from git?

heftig:
I'm using Git and am already patching this downstream, so no urgency from me.

PR #167, opened 2022-11-24
templates/power_profiles: Fix Actions property type

jadahl:
It should be an 'as' (array of strings), not an 'a(as)' (array of tuples containing an array of strings) which it is now.

For reference:

https://gitlab.freedesktop.org/hadess/power-profiles-daemon/-/blob/0.12/src/net.hadess.PowerProfiles.xml#L127


martinpitt (approved):
Whoops, thanks for spotting!

PR #169, opened 2023-01-01
Support loading templates from XDG_DATA_DIRS

### LINKED ISSUE #168: Support loading templates from a shared location
It would be useful if there could be a well-defined location to load templates from, which third party projects could install their templates to without having to submit them as PRs here. That would remove some maintenance burden from python-dbusmock itself, and remove a barrier to using it for third party projects which are used by other third party projects and hence can’t just easily use an in-tree template.

Specifically, I’d like to add a template for eos-updater for consumption by gnome-software, but it seems a bit niche to submit to python-dbusmock.git. So I’d like to be able to install it to (say) `/usr/share/python-dbusmock/templates/eos_updater.py` and have `python3 -m dbusmock -t eos_updater` find it.

I don’t think this is supported already? Thanks

martinpitt:
This is useful for shipping templates in distro -devel packages which
are not generic or stable enough to be maintained in the dbusmock
project.

Fixes #168

----

[rendered documentation](https://github.com/martinpitt/python-dbusmock/tree/xdg-templates#templates)

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,10 +227,24 @@ methods and the standard [dbus-python](https://dbus.freedesktop.org/doc/dbus-pyt
 [`@dbus.service.method`](https://dbus.freedesktop.org/doc/dbus-python/dbus.service.html#dbus.service.method)
 decorator.
 
-If you want to contribute a template, look at
-dbusmock/templates/upower.py for a real-life implementation. You can
-copy dbusmock/templates/SKELETON to your new template file name and
-replace `CHANGEME` with the actual code/values.
+To build your own template, you can copy
+[dbusmock/templates/SKELETON](./dbusmock/templates/SKELETON) to your
+new template file name and replace `CHANGEME` with the actual code/values.
+Look at [dbusmock/templates/upower.py](./dbusmock/templates/upower.py) for
+a real-life implementation.
+
+A template can be loaded from these locations:
+
+ * Provide a path to its `.py` file. This is intended for running tests out of
+   git/build trees with very project specific or unstable templates.
+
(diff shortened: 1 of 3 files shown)


martinpitt:
@pwithnall WDYT?

martinpitt:
@pwithnall : Friendly ping. Is that what you had in mind? No need to review the code, just the README. Thanks!

README.md
```diff
@@ -227,10 +227,24 @@ methods and the standard [dbus-python](https://dbus.freedesktop.org/doc/dbus-pyt
...
+new template file name and replace `CHANGEME` with the actual code/values.
+Look at [dbusmock/templates/upower.py](./dbusmock/templates/upower.py) for
+a real-life implementation.
+
+A template can be loaded from these locations:
+
+ * Provide a path to its `.py` file. This is intended for running tests out of
+   git/build trees with very project specific or unstable templates.
+
+ * From [`XDG_DATA_DIRS/python-dbusmock/templates/`*name*`.py`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
```
pwithnall, line 241:
```suggestion
 * From [`$XDG_DATA_DIRS/python-dbusmock/templates/`*name*`.py`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
```

pwithnall (commented):
This looks great, thanks very much for putting it together.

Sorry for the delay in looking at it, I’ve been swamped. I’ve only got one minor suggestion, though, for what it’s worth.

README.md
martinpitt: Applied, thanks!

martinpitt:
@pwithnall Thanks for checking!

PR #176, opened 2023-06-22
testcase: Wait for PID after killing it

3v1n0:
dbusmock may leave dbus-daemon zombie processes because nothing is
waiting for them, so use waitpid() to ensure we don't end up in such
state

/cc @jadahl


martinpitt (approved):
Thanks a lot!

Issue #177, opened 2023-07-23
powerprofilectl started to SIGTRAP in Fedora Rawhide


martinpitt:
The [recent runs](https://github.com/martinpitt/python-dbusmock/actions/runs/5618534753/job/15238470263) started to fail like this:
```
test_list_holds (tests.test_power_profiles_daemon.TestPowerProfilesDaemon.test_list_holds)
Test holds ... 
(process:714): GLib-CRITICAL **: 14:41:31.101: g_main_context_release() called on a context (0x564d6af1c0d0, owner 0x564d6af16eb0, owner count 3) which is not acquired by the current thread
ERROR
[...]
======================================================================
ERROR: test_list_holds (tests.test_power_profiles_daemon.TestPowerProfilesDaemon.test_list_holds)
Test holds
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/source/tests/test_power_profiles_daemon.py", line 80, in test_list_holds
    out = self.run_powerprofilesctl_list_holds()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/tests/test_power_profiles_daemon.py", line 73, in run_powerprofilesctl_list_holds
    return subprocess.check_output(['powerprofilesctl', 'list-holds'],
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['powerprofilesctl', 'list-holds']' died with <Signals.SIGTRAP: 5>.
```

The other two tests fail in the same way. It's a race condition, it doesn't always happen. E.g. in [this run](https://github.com/martinpitt/python-dbusmock/actions/runs/5618534753/job/15224278017), one test passed and the other two failed.

[...]

PR #178, opened 2023-07-26
spec: Update License: to SPDX format

martinpitt:


martinpitt:
Ignoring rawhide failure, that's tracked in #177

PR #180, opened 2023-08-02
Add initial pytest support

martinpitt:
Add a new dbusmock.pytest_fixtures module which exports an initial set of fixtures. These mostly just wrap DbusTestCase into a fixture, plus some convenience functions for starting a system or session bus. In the future we may have more fixtures to spawn templates, etc.

Ensure that all of our unit tests run with pytest. As that is expensive, only run it for TEST_CODE=1 in Fedora stable. On the others, only run the basic TestAPI checks and the fixtures checks.

This was co-developed with Peter Hutterer @whot, thank you!

---

I took some inspiration from https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/243 and https://github.com/flatpak/xdg-desktop-portal/pull/1062

 - [x] fix unit tests
 - [x] implicitly do the template process cleanup (see below)
 - [x] add documentation to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,8 @@ When using a local system/session bus, you can do unit or integration
 testing without needing root privileges or disturbing a running system.
 The Python API offers some convenience functions like
 `start_session_bus()` and `start_system_bus()` for this, in a
-`DBusTestCase` class (subclass of the standard `unittest.TestCase`).
+`DBusTestCase` class (subclass of the standard `unittest.TestCase`) or
+alternatively as a `@pytest.fixture`.
 
 You can use this with any programming language, as you can run the
 mocker as a normal program. The actual setup of the mock (adding
@@ -44,7 +45,7 @@ objects, methods, properties, and signals) all happen via D-Bus methods
 on the `org.freedesktop.DBus.Mock` interface. You just don't have the
 convenience D-Bus launch API that way.
 
-## Simple example in Python
+## Simple example using Python's unittest
 
 Picking up the above example about mocking upower's `Suspend()` method,
 this is how you would set up a mock upower in your test case:
@@ -124,6 +125,80 @@ Let's walk through:
      machine suspend. Our mock process will log the method call
      together with a time stamp; you can use the latter for doing
      timing related tests, but we just ignore it here.
+
+## Simple example using pytest
+
+The same functionality as above but instead using the pytest fixture provided
+by this package.
+
+```python
+import subprocess
+
+import dbus
+import pytest
+
+import dbusmock
+
+
+@pytest.fixture
+def upower_mock(dbusmock_system):
+    p_mock = dbusmock_system.spawn_server(
+        'org.freedesktop.UPower',
+        '/org/freedesktop/UPower',
+        'org.freedesktop.UPower',
+        system_bus=True,
+        stdout=subprocess.PIPE)
+
+    # Get a proxy for the UPower object's Mock interface
+    dbus_upower_mock = dbus.Interface(dbusmock_system.get_dbus(True).get_object(
+        'org.freedesktop.UPower',
+        '/org/freedesktop/UPower'
+    ), dbusmock.MOCK_IFACE)
+    dbus_upower_mock.AddMethod('', 'Suspend', '', '', '')
+
+    yield p_mock
+
+    p_mock.stdout.close()
+    p_mock.terminate()
+    p_mock.wait()
+
+
+def test_suspend_on_idle(upower_mock):
+    # run your program in a way that should trigger one suspend call
+
+    # now check the log that we got one Suspend() call
+    assert upower_mock.stdout.readline() == b'^[0-9.]+ Suspend$'
+```
+
+Let's walk through:
+
+- We import the `dbusmock_system` fixture from dbusmock which provides us
+  with a system bus started for our test case wherever the
+  `dbusmock_system` argument is used by a test case and/or a pytest
+  fixture.
+
+- The `upower_mock` fixture spawns the mock D-Bus server process for an initial
+  `/org/freedesktop/UPower` object with an `org.freedesktop.UPower`
+  D-Bus interface on the system bus. We capture its stdout to be
+  able to verify that methods were called.
+
+  We then call `org.freedesktop.DBus.Mock.AddMethod()` to add a
+  `Suspend()` method to our new object to the default D-Bus
+  interface. This will not do anything (except log its call to
+  stdout). It takes no input arguments, returns nothing, and does
+  not run any custom code.
+
+  This mock server process is yielded to the test function that uses
+  the `upower_mock` fixture - once the test is complete the process is
+  terminated again.
+
+- `test_suspend_on_idle()` is the actual test case. It needs to run
+  your program in a way that should trigger one suspend call. Your
+  program will try to call `Suspend()`, but as that's now being
+  served by our mock instead of upower, there will not be any actual
+  machine suspend. Our mock process will log the method call
+  together with a time stamp; you can use the latter for doing
+  timing related tests, but we just ignore it here.
 
 ## Simple example from shell
 
@@ -301,9 +376,7 @@ the mock server as a program.
 
 python-dbusmock is hosted on https://github.com/martinpitt/python-dbusmock
 
-Run the unit tests with
-
-    python3 -m unittest
+Run the unit tests with `python3 -m unittest` or `pytest`.
 
 In CI, the unit tests run in containers. You can run them locally with e.g.
 
--- a/dbusmock/pytest_fixtures.py
+++ b/dbusmock/pytest_fixtures.py
@@ -1,0 +1,43 @@
+'''pytest fixtures for DBusMock'''
+
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 3 of the License, or (at your option) any
+# later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
+# of the license.
+
+__author__ = 'Martin Pitt'
+__copyright__ = '(c) 2023 Martin Pitt <martin@piware.de>'
+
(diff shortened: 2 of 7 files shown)


tests/test_pytest.py
```diff
@@ -0,0 +1,29 @@
...
+    p_mock = dbusmock_testcase_session.spawn_server(
+        'org.freedesktop.Test', '/', test_iface, stdout=subprocess.DEVNULL)
+
+    obj_test = dbusmock_testcase_session.get_dbus().get_object('org.freedesktop.Test', '/')
+
+    obj_test.AddMethod('', 'Upper', 's', 's', 'ret = args[0].upper()', interface_name=dbusmock.MOCK_IFACE)
+    assert obj_test.Upper('hello', interface=test_iface) == 'HELLO'
+
+    p_mock.terminate()
+    p_mock.wait()
```
martinpitt, line 18: This sucks. While it passes without the cleanup, that'd be relying too much on process exit behaviour, and may leak too many resources. Ideally spawn_server() can clean this up by itself, but it's class-scoped, so no `addCleanup()`.

I deliberately want to keep the fixtures session-scoped, as opposed to starting a new D-Bus for each single test (which is slow and noisy).

martinpitt:
@whot : this still needs work (see laundry list in the description), but do you have some initial thoughts here?

dbusmock/pytest_fixtures.py
```diff
@@ -0,0 +1,41 @@
...
+    yield testcase
+    testcase.tearDownClass()
+
+
+@pytest.fixture(scope='session')
+def dbusmock_testcase_system(dbusmock_testcase) -> Iterator[dbusmock.testcase.DBusTestCase]:
+    '''Export the whole DBusTestCase as a fixture, with system bus started'''
+
+    dbusmock_testcase.start_system_bus()
+    yield dbusmock_testcase
```
martinpitt, line 33: need to learn/test: maybe `return` here?

whot: yes, `yield` is only useful if you want to do something afterwards, otherwise a return is good enough (which will also drop the `Iterator` from the type annotation.

tests/run-fedora
```diff
@@ -23,12 +23,17 @@ mkdir -p /run/systemd/system
...
-  exit 1
-}
+
+python3 -m unittest -v
+
+# run all tests with pytest only on TEST_CODE, as that is expensive
+if [ -n "$TEST_CODE" ]; then
+    python3 -m pytest -v
+else
+    python3 -m pytest -vv -k 'test_pytest or TestAPI'
```
whot, line 37: fwiw, if you install `python3-pytest-xdist` you can run with `pytest -n auto ...` and it'll run the tests in parallel to speed things up.

tests/test_pytest.py
whot: I think what you want in most cases here is abstract the mock through a local fixture, see #182 where I did this - the tests then use that mock as argument and run the actual test, you need less in-test teardown.

At least that's the best use I've found for the xdg-desktop-portal tests.

whot (commented):
A few minor comments, my bigger comment is effectively #182 which I had mostly ready and just pushed so you can have a look at that approach and we can figure out what the best middle ground is :)

dbusmock/pytest_fixtures.py
martinpitt: Fixed.

tests/test_pytest.py
martinpitt: Thanks! This is nicer indeed, done in one test case to illustrate it.

martinpitt:
Hello @whot  - I'm back from my vacation. Thanks for #182! I took your better ideas from there, fixed the system → session bus confusion in the README, and did some simplifications. I also kept my different file name for the fixtures, as `pytest.py` may be too confusing with `import pytest`.

WDYT?

tests/test_api_pytest.py
```diff
@@ -0,0 +1,36 @@
+import subprocess
+import tempfile
+
+import pytest
+
+import dbusmock
```
whot, line 6:
```suggestion
import dbusmock
import pytest
import subprocess
import tempfile
```
also I just realised this should probably have a copyright header in it like other files, feel free to copy/paste whatever's in the rest of the repo (and this goes for any of my future changes in this repo too).

tests/run-fedora
```diff
@@ -19,12 +19,17 @@ mkdir -p /run/systemd/system
...
+su -s /bin/sh - build << EOF || { [ -z "$DEBUG" ] || sleep infinity; exit 1; }
 set -ex
 cd /source
 export TEST_CODE="$TEST_CODE"
-python3 -m unittest -v || {
-  [ -z "$DEBUG" ] || sleep infinity
-  exit 1
-}
+
+python3 -m unittest -v
```
whot, line 27: How about something like this:
```sh
case "USE_SUITE" in
   unittest)
       RUN_UNITEST=1
       ;;
   pytest)
       RUN_PYTEST=1
       ;;
   *)
       RUN_PYTEST=1
       RUN_UNITTEST=1
       ;;
esac
if [ -n "$RUN_UNITEST" ]; then
   python3 -m unittest -v
fi
if [ "$RUN_PYTEST" ]; then
   python3 -m pytest -v
fi
```
and then in the github workload you can 
```yaml
strategy:
      matrix:
        USE_SUITE: ["unittest", "pytest"]
```
to get the jobs to split up nicely - the default behaviour of the script is to run both.

dbusmock/pytest_fixtures.py
```diff
@@ -0,0 +1,43 @@
...
+    testcase = dbusmock.testcase.DBusTestCase()
+    testcase.setUp()
+    yield testcase
+    testcase.tearDown()
+    testcase.tearDownClass()
+
+
+@pytest.fixture(scope='session')
+def dbusmock_testcase_system(dbusmock_testcase) -> dbusmock.testcase.DBusTestCase:
+    '''Export the whole DBusTestCase as a fixture, with system bus started'''
```
whot, line 32:
```suggestion
    '''Export the whole DBusTestCase as a fixture, with the system bus started'''
```

dbusmock/pytest_fixtures.py
```diff
@@ -0,0 +1,43 @@
...
+
+    dbusmock_testcase.start_system_bus()
+    return dbusmock_testcase
+
+
+@pytest.fixture(scope='session')
+def dbusmock_testcase_session(dbusmock_testcase) -> dbusmock.testcase.DBusTestCase:
+    '''Export the whole DBusTestCase as a fixture, with session bus started'''
```
whot, line 40:
```suggestion
    '''Export the whole DBusTestCase as a fixture, with the session bus started'''
```

whot (approved):
I admit I'm not a huge fan of the "dbusmock_testcase" fixture names (too long) but I can't really come up with something of equal nonambiguity.

A few minor style changes, otherwise this LGTM, thanks!

README.md
```diff
@@ -36,15 +36,15 @@ When using a local system/session bus, you can do unit or integration
...
+alternatively as a `@pytest.fixture`.
 
 You can use this with any programming language, as you can run the
 mocker as a normal program. The actual setup of the mock (adding
 objects, methods, properties, and signals) all happen via D-Bus methods
 on the `org.freedesktop.DBus.Mock` interface. You just don't have the
 convenience D-Bus launch API that way.
-
-## Simple example in Python
+## Simple example using Python's unittest
```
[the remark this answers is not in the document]
martinpitt: Whoops, thanks for spotting! Applied.

README.md
```diff
@@ -125,6 +125,80 @@ Let's walk through:
...
+The same functionality as above but instead using the pytest fixture provided
+by this package.
+
+```python
+import subprocess
+
+import dbus
+import pytest
+
+import dbusmock
```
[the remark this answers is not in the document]
martinpitt: I like to group imports into blocks "stdlib - external modules - internal modules".

tests/test_api_pytest.py
martinpitt: Thanks, added.

dbusmock/pytest_fixtures.py
martinpitt: Applied, thanks!

dbusmock/pytest_fixtures.py
martinpitt: Applied.

tests/run-fedora
martinpitt: TBH I think this is a bit overkill. I dropped the TEST_CODE dependency, and just run everything everywhere. It's not that expensive, and it's good to know when e.g. pytest fails in rawhide.

martinpitt:
> I admit I'm not a huge fan of the "dbusmock_testcase" fixture names (too long) but I can't really come up with something of equal nonambiguity.

Same here. I think it would be okay to shorten them to `dbusmock_{test,system,session}`, WDYT?

whot:
yeah, good idea, I think that's good enough

martinpitt:
@whot Updated. Do you see anything else, or good enough to go? I'm happy with it now.

README.md
```diff
@@ -125,6 +126,80 @@ Let's walk through:
...
+def test_suspend_on_idle(upower_mock):
+    # run your program in a way that should trigger one suspend call
+
+    # now check the log that we got one Suspend() call
+    assert upower_mock.stdout.readline() == b'^[0-9.]+ Suspend$'
+```
+
+Let's walk through:
+
+- We import the `system_mock` fixture from dbusmock which provides us
```
whot, line 175:
```suggestion
- We import the `dbusmock_system` fixture from dbusmock which provides us
```

whot (approved):
one typo left but otherwise thumbs up, thanks!

README.md
martinpitt: Thanks! Fixed.

martinpitt:
Cheers, thanks a lot for this!

PR #182, opened 2023-08-11
Add a pytest module to provide a simple fixture

whot:
@martinpitt: I had this (mostly) ready before I noticed the emails for #180 in my overflowing-after-a-vacation inbox. For easier comparison/review, let's just file a PR and then go from there. It's not 100% ready and tested, but this way we can compare the two PRs and find the best solution.


martinpitt:
Thanks! In #180 I tried to merge your and my original approaches.

martinpitt:
We are converging in #180. Thanks!

PR #183, opened 2023-11-03
If we don't have a child to wait for, stop

whot:
No need to try to os.kill() that same pid if it no longer exists. Worst case we may end up killing an unrelated process.


martinpitt (approved):
Nice catch, thanks!

PR #184, opened 2023-11-06
Fix inheritance issues and separate the mocked dbus into an object

whot:
Filing as draft for now because the last commit *may* break backwards compat, see below.

I noticed that my dbus-daemons didn't get terminated when using dbusmock in the xdg-desktop-portal which led to a bit of a rabbit hole. So I figured "let's fix this properly" :smile: 

For (understandable) historical reasons, the `DBusTestCase` class handled everything as `@classmethod`. That causes a few issues though because the `cls` argument is mostly the real class but in some cases it's `DBusTestCase`. This can lead to a getter resolving to`DBusTestCase.somevar` but the setter resolving to `MyTestClass(DBusTestCase).somevar`.

The first few patches disentangle that to make sure we're always on the correct class instead of oscillating between `DBusTestCase` and the test class.

The last (and most interesting) patch is to add a new `MockedDBus` object that wraps the actually started daemon.

I've patched this so `DBusTestCase` wraps this new class but this *may* introduce some compatibility issues. Arguably we could leave `DBusTestCase` untouched and as legacy API and suggest using `MockedDBus` via pytest for the future. Since the pytests haven't been in a release yet we could also remove some of the wrappers from this new class and rely on callers to use the new `BusType`/`spawn_*` API directly.

Ironically, the last patch somewhat undoes the previous patches in that everything is now on the `DBusTestCase` class, but that shouldn't matter much.

It does pass the test suite here, so any incompatibilities introduced are not tested ;)

### DIFF
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -66,12 +66,11 @@ class BusType(enum.Enum):
         dbus_if = dbus.Interface(dbus_obj, 'org.freedesktop.DBus')
         dbus_if.ReloadConfig()
 
-    def wait_for_bus_object(self, dest: str, path: str, timeout: int = 600):
+    def wait_for_bus_object(self, dest: str, path: str, timeout: float = 60.0):
         '''Wait for an object to appear on D-Bus
 
         Raise an exception if object does not appear within one minute. You can
-        change the timeout with the "timeout" keyword argument which specifies
-        deciseconds.
+        change the timeout in seconds with the "timeout" keyword argument.
         '''
         bus = self.get_connection()
 
@@ -91,7 +90,7 @@ class BusType(enum.Enum):
                     if '.UnknownInterface' in str(e):
                         break
 
-            timeout -= 1
+            timeout -= 0.1
             time.sleep(0.1)
         if timeout <= 0:
             assert timeout > 0, f'timed out waiting for D-Bus object {path}: {last_exc}'
@@ -380,7 +379,7 @@ class DBusTestCase(unittest.TestCase):
         BusType.wait_for_bus_object() instead.
         '''
         bustype = BusType.SYSTEM if system_bus else BusType.SESSION
-        bustype.wait_for_bus_object(dest, path, timeout)
+        bustype.wait_for_bus_object(dest, path, timeout / 10.0)
 
     @staticmethod
     def spawn_server(name: str, path: str, interface: str, system_bus: bool = False, stdout=None) -> subprocess.Popen:
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -110,7 +110,8 @@ class DBusTestCase(unittest.TestCase):
 
         This gets stopped automatically at class teardown.
         '''
-        DBusTestCase.__start_bus('session')
+        assert cls.session_bus_pid is None
+        cls.__start_bus('session')
 
     @classmethod
     def start_system_bus(cls) -> None:
@@ -118,7 +119,8 @@ class DBusTestCase(unittest.TestCase):
 
         This gets stopped automatically at class teardown.
         '''
-        DBusTestCase.__start_bus('system')
+        assert cls.system_bus_pid is None
+        cls.__start_bus('system')
 
     @classmethod
     def start_dbus(cls, conf: Optional[str] = None) -> Tuple[int, str]:
--- a/dbusmock/testcase.py
+++ b/dbusmock/testcase.py
@@ -26,6 +26,7 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Tuple
 
 import dbus
+import dbus.proxies
 
 from dbusmock.mockobject import MOCK_IFACE, OBJECT_MANAGER_IFACE, load_module
 
@@ -234,10 +235,11 @@ class PrivateDBus:
         if self._daemon:
             self.bustype.reload_configuration()
 
-    def spawn_server(self, name: str, path: str, interface: str, system_bus: bool = False, stdout=None):
+    def spawn_server(self, name: str, path: str, interface: str, system_bus: bool = False, stdout=None) -> subprocess.Popen:
         '''Wrapper around ``spawn_server`` for backwards compatibility'''
         assert not system_bus or self.bustype == BusType.SYSTEM, "Mismatching bus types"
-        return spawn_server(name=name, path=path, interface=interface, bustype=self.bustype, stdout=stdout)
+        server = SpawnedMock.spawn_for_name(name, path, interface, bustype=self.bustype, stdout=stdout)
+        return server.process
 
     def wait_for_bus_object(self, dest: str, path: str, system_bus: bool = False, timeout: int = 600):
         '''Wrapper around ``BusType.wait_for_bus_object()`` for backwards compatibility'''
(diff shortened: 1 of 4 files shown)


dbusmock/testcase.py
```diff
@@ -28,6 +30,260 @@
...
+
+    @property
+    def pid(self) -> int:
+        assert self._daemon is not None, "Call start() first"
+        return self._daemon.pid
+
+    def start(self):
+        '''Start the D-Bus daemon.'''
+        addr = f'unix:path={self._socket}'
+        argv = ['dbus-daemon', f'--address={addr}', f'--config-file={self._config}']
```
whot, line 190: actually, I'm pretty sure we don't need the `--address` arg here since we specify it in the given config file and `--address` is only needed to override that.

whot:
ftr I'm also planning to move `spawn_*` into a context manager objects so we can do things like:
```python
with spawn_server() as blah:
   .... 
```
and have it auto-terminate instead of requiring callers to track the Popen objects.

dbusmock/testcase.py
```diff
@@ -63,9 +63,9 @@ def get_services_dir(cls, system_bus: bool = False) -> str:
 
     @classmethod
     def tearDownClass(cls):
-        if cls._DBusTestCase__datadir:
-            shutil.rmtree(cls._DBusTestCase__datadir)
-            cls._DBusTestCase__datadir = None
+        if DBusTestCase._DBusTestCase__datadir:
```
martinpitt: Oh yes, well spotted -- that really wants to be a global variable, back then I wasn't aware that `cls` would actually be a different object for a derived class.

dbusmock/testcase.py
```diff
@@ -113,15 +113,17 @@ def start_session_bus(cls) -> None:
 
         This gets stopped automatically at class teardown.
         '''
-        DBusTestCase.__start_bus('session')
+        assert cls.session_bus_pid is None
+        cls.__start_bus('session')
```
martinpitt: Starting a bus per (derived) class feels oddly in the middle between concepts: Either every test needs its own bus (and then this shouldn't be a classmethod - but that feels a bit excessive), or all test classes are fine with sharing a single bus; the latter was actually the intention, as this starting/stopping happens per test file. Your assertion also seems to support this. 

So I'd actually be in favor of fixing it the other way round: Leaving start_ as it is (global singleton), and fixing the teardown instead. WDYT?

dbusmock/testcase.py
```diff
@@ -106,6 +106,8 @@ def __start_bus(cls, bus_type) -> None:
         (pid, addr) = cls.start_dbus(conf=str(conf))
         os.environ[f'DBUS_{bus_type.upper()}_BUS_ADDRESS'] = addr
         setattr(cls, f'{bus_type}_bus_pid', pid)
+        assert DBusTestCase.session_bus_pid is None
+        assert DBusTestCase.system_bus_pid is None
```
martinpitt: See my previous comment. If we go back to actual global d-bus instances, these should be inverted to `is not`.

dbusmock/testcase.py
```diff
@@ -28,6 +30,260 @@
...
+logger = logging.getLogger("dbusmock")
+logger.setLevel(logging.DEBUG)
+
+
+class BusType(enum.Enum):
+    SESSION = "session"
+    SYSTEM = "system"
+
+    @staticmethod
+    def from_bool(system_bus: bool) -> "BusType":
```
martinpitt, line 42: Why the quoted return type? Doesn't c8s's python 3.6 understand an unquoted class name yet?

dbusmock/testcase.py
```diff
@@ -28,6 +30,260 @@
...
+                    if '.UnknownInterface' in str(e):
+                        break
+
+            timeout -= 1
+            time.sleep(0.1)
+        if timeout <= 0:
+            assert timeout > 0, f'timed out waiting for D-Bus object {path}: {last_exc}'
+
+
+class MockedDBus:
```
martinpitt, line 115: Bikeshedding: This isn't really a "mocked" bus -- it's a fully functional real dbus-daemon/dbus-broker.  The point of it to be isolated. So perhaps {Test,Private,Isolated,Sandbox}DBus?

dbusmock/testcase.py
```diff
@@ -28,6 +30,260 @@
...
+    '''A D-Bus daemon instance that prepresents either a session or a system bus.
```
martinpitt, line 116: "p"represents typo. I don't mind the various typos in the commit messages much, but it's more important in the docstrings.

dbusmock/testcase.py
```diff
@@ -41,81 +297,57 @@ class DBusTestCase(unittest.TestCase):
...
+    }
+
+    @staticmethod
+    def _bus(bustype: BusType) -> MockedDBus:
+        '''
+        Return (and create if necessary) the singleton DBus for the given bus type.
+        '''
+        if not DBusTestCase._busses.get(bustype):
+            DBusTestCase._busses[bustype] = MockedDBus(bustype)
+        return DBusTestCase._busses.get(bustype)  # type: ignore
```
martinpitt, line 312: This can be a normal `[bustype]` instead of .get(), as it really ought to exist at this point.

martinpitt (changes requested):
Thanks @whot for your work here! I like this. It's still draft and there are several test failures, but I did a first round of review to get some feedback.

Sorry about the conflict -- I just pushed a branch to fix the CI regression on main, which happened because fedora:latest moved from F38 to F39.

dbusmock/testcase.py
whot: The type isn't available until it's fully defined and we're inside `BusType` here, so it's not yet defined at this point. But after a bit of googling this should be `-> Self` instead.

dbusmock/testcase.py
whot: I'm in two minds about this. And there are two independent issues with this, I think.

A global singleton would be simpler but this is a testcase class wrapper so I'd guess you want sets of tests (i.e. a class) isolated from each other and not interfere with other sets. Right now if you run test classes in parallel in the same process one class's start/stop would override the other class' bus (Not sure it will actually happen and how you'd trigger this in the same process). But if everything uses the same singleton it can happen. Plus you overwrite the pid so you can't actually stop andy but the most recently started bus. So that would favour the per-class bus.

However: we rely on `setenv`/`getenv` so some  interference and race condition is somewhat guaranteed anyway. Then again that's only on startup and only a narrow window compared to the start/stop sequence.

Functionally, nothing actually changes here because the way unittest and pytest call the fixtures you'll always call `tearDownClass()` first which will stop the current bus. So I think it doesn't matter where the bus is attached to.

Lots of words to say :shrug:, dunno. IMO the right way here is to push consumers towards pytest fixtures instead :)

whot (commented):
Thx for the review, much appreciated. I'll get this updated next week.

dbusmock/testcase.py
whot: Turns out `Self` is not available until [python 3.11](https://peps.python.org/pep-0673/), so I'd say let's keep the quoted one.

martinpitt: Ah, ok. We can clean these up once RHEL 8 goes into maintenance mode.

dbusmock/testcase.py
martinpitt: Good point about the tearDownClass(). Let's keep this approach then.

dbusmock/testcase.py
whot: removed now

dbusmock/testcase.py
whot: Renamed to `PrivateDBus`. fwiw, I think we should/could move this into dbusmock proper - while it's only used by the tests there is no longer anything test-specific about it.

dbusmock/testcase.py
whot: I think I fixed them all now, or at least I've stared at it for long enough that I no longer see them...

whot:
Updated, with the main change aside from the various fixes being that there's now a `SpawnedServer` object that provides a context manager for the spawn_server helpers - easier to use and doesn't require the caller to keep track of subprocesses. I don't like the name, can't think of anything better though.

dbusmock/testcase.py
whot: Resolving this as "fixed" or rather "let's leave it", see [the thread in the previous comment](https://github.com/martinpitt/python-dbusmock/pull/184#discussion_r1388976334)

whot:
Ok, pipeline is green and I think this is ready. Or at least I've rebased and stared at it enough that I no longer see any bugs...

martinpitt (approved):
This looks great to me, thanks a lot for your work here!

martinpitt:
@whot Do you want this in a release now, or do you have some more things you want to work on before that?

whot:
I have nothing else in the pipe right now (beyond whatever bugs I just introduced in this MR ;)

Thanks heaps for the quick merge!

martinpitt:
Ack, here is https://github.com/martinpitt/python-dbusmock/releases/tag/0.30.0 . Uploaded to Debian, and Fedora updates are in progress.

PR #185, opened 2023-11-07
doc: hook up sphinx

whot:
For reference, this is how it looks like: https://whot.github.io/python-dbusmock/
Needs enabling in https://github.com/martinpitt/python-dbusmock/settings/pages, change to `Source: Github Action`.

This is the minimal conversion only, using the readme as index page and with a few rst-ifications in the `DBusMockObject` class (but not in `DBusTestCase` because that would conflict with #184). Once it's in place we can prettify all this one-by-one.

~~A helper `Makefile` is included too to build it locally.~~

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,0 +1,49 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'python-dbusmock'
+copyright = '2023, Martin Pitt'  # noqa: A001
+author = 'Martin Pitt'
+
+try:
+    # created by setuptools_scm
+    from dbusmock._version import __version__ as release
+except ImportError:
+    release = '0.git'
+
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx.ext.viewcode',
+    'sphinx.ext.intersphinx',
+    'sphinx.ext.napoleon',
+    'myst_parser',
+    'autoapi.extension',
+]
+
+templates_path = ['_templates']
+exclude_patterns = []
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,18 @@
+# Minimal makefile for Sphinx documentation
```
martinpitt, line 1: I really want to avoid a top-level Makefile. This is a Python project, and this may even confuse build system autodetection for distributions/packagers, let alone humans. If there is no pyproject.toml moral equivalent to npm package.json's `"scripts:" {}`, then perhaps a `./build-sphinx.sh`?

whot: The Makefile is just a helper for us (dbusmock devs), even the CI just calls `sphinx-build -M html doc/ doc/` directly. I've dropped it now.

I'm not sure how much value a locally built sphinx-build is over building it online and referring to readthedocs. So I'm not sure integrating it into `pyproject.toml` is really worth it? If anything we'll just have to spend time debugging it where it doesn't work :)

doc/conf.py
```diff
@@ -0,0 +1,44 @@
...
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'python-dbusmock'
+copyright = '2023, Martin Pitt'  # noqa: A001
+author = 'Martin Pitt'
+release = '0.29.1'
```
martinpitt, line 12: This needs to use the dynamic version like [here](https://github.com/martinpitt/python-dbusmock/blob/main/dbusmock/__init__.py#L20).

martinpitt: Hmm, does this even appear anywhere? I don't see it on https://whot.github.io/python-dbusmock/index.html or https://whot.github.io/python-dbusmock/autoapi/dbusmock/index.html or https://whot.github.io/python-dbusmock/search.html?q=0.29.1

whot: I had to grep the source code and yes, it does show up:
`  <title>python-dbusmock &mdash; python-dbusmock 0.29.1 documentation</title>`

Guess with everyone having a thousand tabs open no-one would notice that ;)

martinpitt: Locally the dynamic import works in the sense of sphinx-build succeeding. I pushed that.

I also added the two new directories to .gitignore.

martinpitt (approved):
Thank you, this is really nice! I'm waiting for CI to complete. I did the project settings change, so this ought to work after landing.

martinpitt:
Et voilà: https://martinpitt.github.io/python-dbusmock/ :rocket:

whot:
it's like the future! :)

PR #186, opened 2023-11-29
Default to running static tests

whot:
unittest and pytest may pass locally only to fail with
near-indecipherable logs in the CI - caused by TEST_CODE being set which
triggers the pylint, ruff and mypy checks.

Let's invert that check so local checks will run these by default unless
explicitly disabled and the CI disables it in the specific test runs.

### DIFF
--- a/tests/run
+++ b/tests/run
@@ -9,10 +9,10 @@ else
 fi
 
 # only run static code checks on a single release, too annoying to keep the code compatible with multiple versions
-if [ "${IMAGE%fedora:latest}" != "$IMAGE" ]; then
-    TEST_CODE="1"
+if [ "${IMAGE%fedora:latest}" = "$IMAGE" ]; then
+    SKIP_STATIC_CHECKS="1"
 fi
 
 OS=${IMAGE##*/}
 OS=${OS%:*}
-$RUNC run --interactive -e DEBUG=${DEBUG:-} -e TEST_CODE="${TEST_CODE:-}" --rm ${RUNC_OPTIONS:-} --volume `pwd`:/source:ro --workdir /source "$IMAGE" /bin/sh tests/run-$OS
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
Sure, thanks!

FTR, I use the vim [ALE plugin](https://github.com/dense-analysis/ale) which immediately shows me mypy/ruff/flake8/eslint/etc. issues inline, so I didn't feel that problem. But of course not everybody has these checkers enabled in their editor.

whot:
Thanks, didn't know about that, I'll have a look. I had whatever is enabled by default in LunarVim, so *some* warnings I saw and fixed as is but I guess I've been too reliant on black for things like spaces :)

Also the problem with any editor plugins is that they may default to a different style then the project (though using `pyproject.toml` should avoid that situation).

martinpitt:
Yes, ALE is pretty great for that -- it reads settings from pyproject.toml, .flake8, .eslintrc.json etc. It doesn't need a lot of config or attention, it basically just DTRT. I have these settings in my .vimrc which tweak it a little, but the defaults are already great:

```
let g:ale_rust_cargo_use_clippy = 1
let g:ale_echo_msg_format = '[%linter%|%severity%] %s'
let g:ale_sign_error = '❌'
let g:ale_sign_warning = '⚠️'
let g:ale_fixers = {
\    '*': ['remove_trailing_lines', 'trim_whitespace'],
\    'rust': ['remove_trailing_lines', 'trim_whitespace', 'rustfmt'],
\    'javascript': ['remove_trailing_lines', 'trim_whitespace', 'eslint'],
\}
let g:ale_fix_on_save = 1
```

whot:
ftr and because i finally found it: lunarvim by default uses [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim) as default plugin. at least now I know what's displaying the various complaints in my buffer :)

PR #188, opened 2023-12-01
Use black as code formatter

### LINKED ISSUE #187: Use black (or blue) for code style?
I just spent way too long chasing down pylint and ruff issues (which were a bit hidden due to the magic env var) so now I'm wondering: would you be amenable to using [black](https://pypi.org/project/black/) (or [blue](https://pypi.org/project/blue/) [^1])?

It's going to be one large commit to reformat everything but in the projects I've used it it's been a blessing for it to take care of almost all ruff/pylint cases - plus the consistent style (even if it's not always perfect)

In the CI it'd require one job to do `black --check --diff .` and fail if that fails. And black can quite easily be integrated into [pre-commit](https://github.com/pre-commit/pre-commit-hooks) as well for example.

[^1]: haven't used blue myself but it's supposed to be more configurable than black.

whot:
Integrated into pyproject.toml and tests/test_code.py so we can enforce the formatting in the CI.

Fixes #187

---
Quick! let's do this before someone has an idea of refactoring everything :)

### DIFF
--- a/dbusmock/__main__.py
+++ b/dbusmock/__main__.py
@@ -1,5 +1,5 @@
 # coding: UTF-8
-'''Main entry point for running mock server.'''
+"""Main entry point for running mock server."""
 
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of the GNU Lesser General Public License as published by the Free
@@ -7,11 +7,11 @@
 # later version.  See http://www.gnu.org/copyleft/lgpl.html for the full text
 # of the license.
 
-__author__ = 'Martin Pitt'
-__copyright__ = '''
+__author__ = "Martin Pitt"
+__copyright__ = """
 (c) 2012 Canonical Ltd.
 (c) 2017 - 2022 Martin Pitt <martin@piware.de>
-'''
+"""
 
 import argparse
 import json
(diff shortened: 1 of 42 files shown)


whot:
For easier review: manual changes in `pyproject.toml`, `tests/test_code.py` (to run black) and `tests/run-fedora` (to install black), rest is whatever black generated though I had to fix a small number of things like `b"foo " b"bar"` one-liners that for some reason didn't resolve to `b"foo bar"`.

tests/test_api.py
```diff
@@ -45,18 +45,20 @@ def setUpClass(cls):
...
         self.dbus_props = dbus.Interface(self.obj_test, dbus.PROPERTIES_IFACE)
 
     def assertLog(self, regex):
-        self.assertRegex(Path(self.mock_log.name,).read_bytes(), regex)
+        self.assertRegex(
+            Path(
+                self.mock_log.name,
+            ).read_bytes(),
+            regex,
+        )
```
martinpitt, line 61: Ugh -- this is "I get paid by written lines of code" vandalism. Why is black so trigger-happy with "put a single word on each line"? This feels quite arbitrary, especially as [further below](https://github.com/martinpitt/python-dbusmock/pull/188/files#diff-3451cc474d5698222168f66a293339a6bd296ee2c56f56409866bbae99917a79L255) it did the exact opposite :thinking: 

After some interactive experimenting: the whole problem is that stray `,` in the Path() call. If you drop this, then black leaves it alone.

dbusmock/__main__.py
```diff
@@ -24,46 +24,54 @@
...
+    parser.add_argument(
+        "--session",
+        action="store_true",
+        help="put object(s) on session bus (default without template; overrides template's SYSTEM_BUS flag)",
+    )
+    parser.add_argument("-l", "--logfile", metavar="PATH", help="path of log file")
+    parser.add_argument("-t", "--template", metavar="NAME", help="template to load (instead of specifying name, path, interface)")
+    parser.add_argument(
+        "name", metavar="NAME", nargs="?", help='D-Bus name to claim (e. g. "com.example.MyService") (if not using -t)'
+    )
```
martinpitt, line 45: This kind of formatting inconsistency makes me weep. Three different styles to format the same kind of function call.. Especially the third one makes no sense at all IMHO. Either everything into one line, or split into one arg per line. I suppose this needs some manual reformatting and arguing until human sense of beauty/readability and black eventually agree?

pyproject.toml
```diff
@@ -88,3 +88,6 @@ ignore = [
     "RUF012", # Mutable class attributes should be annotated with `typing.ClassVar`
     "SIM105", # Use `contextlib.suppress(KeyError)` instead of `try`-`except`-`pass`
 ]
+
+[tool.black]
+line-length = 130
```
martinpitt, line 93: As I said, my target is 118. But starting with 130 here is fine, then we can send a follow-up that changes it to 118. (FTR: I ran `black --line-length=118 --check --diff .|view -` locally against this PR, and that relative diff looks sensible).

martinpitt (changes requested):
Thanks! This looks 95% good, just needs a little manual work for the argparse code.

dbusmock/__main__.py
whot: unfortunately black isn't smart enough to know they're the same calls so it just goes by whether stuff fits on a line. In the first case the three args don't fit on one line within the line break which is why it's broken up in three different ways.

Looks like the way around this (sort-of) is to add a stray `,` after the help text and then each arg will be on one line. That blows out the code a bit though. Alternatively: `# fmt: off` works too for specific cases where we really don't want black to touch it.

martinpitt: The change to 118 char lines actually helped to make this more consistent.

pyproject.toml
martinpitt: Thanks! Looks good IMHO.

dbusmock/__main__.py
martinpitt: Meh, not important enough. Let's fix pylint and be done with it.

whot: there's basically zero wriggle room, only whatever `black --help` suggests which is... line length and string quotes (apparently? that must be new-ish).

whot:
fixed the pylint error with a follow-up to make it unnecessary

martinpitt (dismissed):
Cheers! Assuming green, this looks good. I'll change the max line length for the other tools as a follow-up.

martinpitt:
Oh, [one more black error](https://github.com/martinpitt/python-dbusmock/actions/runs/7057344106/job/19210855133?pr=188#step:3:907)

whot:
>  I'll change the max line length for the other tools as a follow-up.

arguably we should leave them as-is. Black will format the lines to 118 and enforce that, if you leave pylint/ruff on 130 (or remove the line length) then we can format with a trailing `# fmt: off` without having those two tools complaining about it (happened to me in the earlier 130 version)

> Oh, [one more black error](https://github.com/martinpitt/python-dbusmock/actions/runs/7057344106/job/19210855133?pr=188#step:3:907)

gah. fixed now (friday evening here, I probably shouldn't be pushing things :)

martinpitt:
Now off to enabling black in vim :grin:

PR #190, opened 2023-12-04
bluez5: Set Modalias, Class & Icon properties

z3ntu:
Otherwise we only emit those with PropertiesChanged but GetAll won't return those properties.

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -516,6 +516,10 @@ def PairDevice(_self, adapter_device_name, device_address, class_):
             },
         )
 
+    device.props[DEVICE_IFACE]['Modalias'] = dbus.String("bluetooth:v000Fp1200d1436", variant_level=1)
+    device.props[DEVICE_IFACE]['Class'] = dbus.UInt32(class_, variant_level=1)
+    device.props[DEVICE_IFACE]['Icon'] = dbus.String("phone", variant_level=1)
+
     device.EmitSignal(
         dbus.PROPERTIES_IFACE,
         "PropertiesChanged",


z3ntu:
Ah, I misread the code a bit.

Seems the `try` block succeeds (at least in my environment) so `AddProperties` is never called and then obviously setting the properties like I'm doing here is the correct thing.

So the only change I could imagine is moving the assignment of the properties into the `try:` block but not sure if it makes anything better so I'm keeping it like now.

martinpitt:
I'm not sure why the AddProperties() call was conditionalized in this form, i.e. only checking for an existing Modalias. @pwithnall added that ten years ago in commit ee29a4403359b6a -- perhaps he still remembers?

But repeating the key/values and unconditionally stomping over previously defined properties is wrong.

`AddProperties()` already updates `self.props`, so this is also redundant. My suggestion is to split the "property already exists" into three independent checks to add each missing property individually.

Thanks!

pwithnall:
> I'm not sure why the AddProperties() call was conditionalized in this form, i.e. only checking for an existing Modalias. @pwithnall added that ten years ago in commit [ee29a44](https://github.com/martinpitt/python-dbusmock/commit/ee29a4403359b6a22027a39a5c39b6e8b6cbdcef) -- perhaps he still remembers?

My guess would be that it was to support the flow of calling `PairDevice` on an existing device to unblock it (as per the new documentation line I added at the top of that commit). The conditionalisation would avoid `Modalias` (etc.) being re-added as properties to a device which already has them (but which was blocked).

That’s a guess, though. I can’t remember what I was doing last week, let alone what I was thinking when writing a commit 10 years ago. :grin:

martinpitt:
We wrote software TEN YEARS AGO!? :grinning: Thanks @pwithnall for the fast response! So as long as you don't remember anything more specific and don't object to checking/adding props one by one, let's go with that.

martinpitt:
@z3ntu conceptually yes, possibly a bit more elegantly with iterating over some `default_props = {...}` dict, but I won't be too picky :grin: Thanks!

z3ntu:
Okay I checked again, and that wouldn't work. The properties *do* exist but they're empty string (Modalias & Icon) or 0 (Class).

But without reading further into how dbusmock works, why can we not set the properties like ~10 lines above where all the other props are set? Why do you not want to set the prop if it already exists?

martinpitt:
@z3ntu I'm afraid I can't do positive definitive claims about any logic here -- this template is the work of @pwithnall and @hadess and I'm afraid I don't know anything about its consumers or the Bluetooth stack. This is what I can say:

> But without reading further into how dbusmock works

No need to. All of the logic is confined to bluez5.py. The generic dbusmock code does the general plumbing around it, but it doesn't touch any of the properties here.

> why can we not set the properties like ~10 lines above where all the other props are set? 

In that block, PairDevice() sets properties that are *supposed* to change after pairing, i. e. Paired=True and a few related ones. That is a legitimate part of the state logic what a real BT device would do. OTOH, at least Modalias is *not* a dynamic property. Pairing a device should not change it. It's less obvious for Icon, but my hunch is that a keyboard stays a keyboard and a headphone stays a headphone, no matter if it's paired or not. These are static properties, and I'm not sure why PairDevice() would ever touch them at all.

Naïvely, the same is true for `Class`  -- however, PairDevice() gets an explicit `class_` parameter. That's the bit where I really need help -- why is that done *there*, as opposed to being a parameter of AddDevice()? This feels like a static property, not one that changes during pairing. This was done in commit fae4be7f49c0861ff by @bjoernh .

> Why do you not want to set the prop if it already exists?

See above -- it seems legit for the caller to modify them after calling AddDevice(), as that only sets them to some constant defaults -- in particular, Modalias/Class/Icon are set to "". Pairing a device cannot suddenly change them.

That original intention of commit ee29a4403359b6a220 was to add these static properties as defaults if they weren't present. But commit 59d6af0dca3ef silently broke this when it made AddDevice() initialize these properties as empty strings.

So my current gut feeling is that this needs some cleanup:

 * Kill all the static device property mangling in PairDevice()
 * Instead, set these values from there in AddDevice() as defaults, instead of empty strings.
 * Revert commit fae4be7f49c0861ff. It was wrong, and I'm sorry that I wasn't more insistent back then. D-Bus methods really should not have default parameters, and I already had to fix it in commit 8968284e8b. Callers can change the Class property after AddDevice(). (I don't want to break that API either).

I'll send a PR, then we can discuss something more concrete, and @z3ntu can check if that unbreaks things for him.

martinpitt:
I sent PR #192 to do what I discussed above. Please have a look, thanks!

PR #191, opened 2023-12-14
README: add a link to the github pages docs

whot:
Some of the documentation should probably move there altogether but for now let's at least have a link to it.


martinpitt (approved):
Danke!

PR #192, opened 2023-12-21
bluez5: Clean up static default properties, re-drop PairDevice class_ parameter

martinpitt:
Over the years, this template has accumulated some hacks and bad API which made PairDevice()'s handling of the Modalias/Class/Icon properties buggy and hard to understand:

 * These are *static* device properties, they are not supposed to change during pairing.
 * Commit ee29a4403359b6a added these as some kind of "dynamic fallback default" when they were not initialized by the caller after AddDevice().
 * Commit 59d6af0dca3e silently broke that fallback default by changing AddDevice() to set these device properties to empty strings.
 * Commit fae4be7f49c0861 added another really bad API for setting Class in PairDevice()(). That API didn't fit into D-Bus (see commit 8968284e8b which had to make it a non-default parameter) and also broke the API, and moreover it is totally unintuitive -- the device class has nothing to do with pairing.

Clean up all of these: Set the static property defaults in AddDevice() right away, so that the caller can adjust them afterwards. Re-drop the `class_` argument in PairDevice(). Adjust the documentation of AddDevice() to point out that properties should be changed after calling that.

Consequently, PairDevice() will stop claiming that the static properties changed. This also gets rid of some redundant code.

-----

@z3ntu, this is meant to fix and Closes #190 .

@pwithnall @hadess @bjoernh I admittedly don't understand the use cases/current callers of this code. I think some API change is inevitable here to clean up, but it should at least make sense. I'd appreciate if you could check my reasoning and do some sanity testing on this with your existing tests. Cheers!

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -383,7 +383,9 @@ def AddDevice(self, adapter_device_name, device_address, alias):
     for the device (e.g. as set on the device itself), and the adapter device
     name is the device_name passed to AddAdapter.
 
-    This will create a new, unpaired and unconnected device.
+    This will create a new, unpaired and unconnected device with some default properties
+    like MOCK_PHONE_CLASS "Class" and a static "Modalias". Especially when working with more
+    than one device, you may want to change these after creation.
 
     Returns the new object path.
     """
@@ -400,8 +402,8 @@ def AddDevice(self, adapter_device_name, device_address, alias):
         "Address": dbus.String(device_address, variant_level=1),
         "AddressType": dbus.String("public", variant_level=1),
         "Name": dbus.String(alias, variant_level=1),
-        "Icon": dbus.String("", variant_level=1),
-        "Class": dbus.UInt32(0, variant_level=1),
+        "Icon": dbus.String("phone", variant_level=1),
+        "Class": dbus.UInt32(MOCK_PHONE_CLASS, variant_level=1),
         "Appearance": dbus.UInt16(0, variant_level=1),
         "UUIDs": dbus.Array([], signature="s", variant_level=1),
         "Paired": dbus.Boolean(False, variant_level=1),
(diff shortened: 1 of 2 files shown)


dbusmock/templates/bluez5.py
```diff
@@ -504,33 +506,13 @@ def PairDevice(_self, adapter_device_name, device_address, class_):
...
-            {
-                "UUIDs": dbus.Array(uuids, variant_level=1),
-                "Paired": dbus.Boolean(True, variant_level=1),
-                "LegacyPairing": dbus.Boolean(True, variant_level=1),
-                "Blocked": dbus.Boolean(False, variant_level=1),
-                "Modalias": dbus.String("bluetooth:v000Fp1200d1436", variant_level=1),
-                "Class": dbus.UInt32(class_, variant_level=1),
-                "Icon": dbus.String("phone", variant_level=1),
-            },
+            device.props[DEVICE_IFACE],
```
martinpitt, line 515: Wait, this is nonsense. Fixing..

martinpitt: Done.

pwithnall:
> I admittedly don't understand the use cases/current callers of this code. I think some API change is inevitable here to clean up, but it should at least make sense. I'd appreciate if you could check my reasoning and do some sanity testing on this with your existing tests. Cheers!

IIRC, I was using BlueZ in python-dbusmock to test part of libfolks, which is something I haven’t worked on for a long time now. I’m afraid I’m not going to resurrect my involvement in it to test this PR, sorry! If (in the unlikely case that) this PR breaks things in libfolks, I’m sure libfolks can be reworked to use the new approach here.

martinpitt:
Thanks @pwithnall. @ricotz and @nielsdg you seem to have worked on libfolks recently -- does this change work for/make sense to you?

nielsdg:
I think it's fine to land this since it seems to fix a couple of issues and we can fix this in a reactive manner once it breaks CI (and honestly, i had to fix our CI for something similar a while ago already, see https://gitlab.gnome.org/GNOME/folks/-/commit/b6b7d41aad9e36e0). The bluez backend is barely used, and this would only affect testing at this point in any case.

martinpitt:
Thanks @nielsdg for checking! Let's go with this then, even if it causes some noise I think it'll make the whole thing a lot easier to understand and adjust at least.

martinpitt:
@nielsdg And right, https://gitlab.gnome.org/GNOME/folks/-/commit/b6b7d41aad9e36e0 would have to be reverted again.

z3ntu:
Sorry for the delay, I tested with lomiri-system-settings tests and they do work also with this PR :) Thanks for fixing this!

Issue #193, opened 2023-12-30
dbusmock/templates/bluez5.py still calls PairDevice with 3 arguments


smcv:
The dbusmock mockup of bluez appears to implement `PairDevice()`, a dbusmock-specific convenience method, which changed from 3 to 2 arguments in 63264e18. Unfortunately, `dbusmock/templates/bluez5.py` still passes the third (class) argument, leading to the gnome-bluetooth test suite failing, even after applying https://gitlab.gnome.org/GNOME/gnome-bluetooth/-/merge_requests/175:

```
1703969200.285 GetManagedObjects
1703969200.285 GetAll /org/freedesktop/UPower org.freedesktop.UPower
1703969200.286 EnumerateDevices
1703969200.287 Pair
1703969200.287 Pair raised: org.freedesktop.DBus.Error.InvalidArgs: Invalid arguments: Fewer items found in D-Bus signature than in Python arguments
```

(I think it would also be good if any dbusmock-specific methods could be renamed to `MockPairDevice()` or similar to make it clearer that they are not part of the real API that is being mocked-up.)

PR #194, opened 2023-12-30
bluez5: Fix invalid arguments to PairDevice

smcv:
The third (device class) argument to PairDevice was removed in 0.30.1, but this call to it was still passing a third parameter, resulting in an error from dbus-python whenever Pair() was called. This caused a unit test regression in gnome-bluetooth.

Fixes: 63264e18 "bluez5: Clean up static default properties, re-drop PairDevice class_ parameter"  
Resolves: https://github.com/martinpitt/python-dbusmock/issues/193  
Bug-Debian: https://bugs.debian.org/1057564


martinpitt (approved):
Argh argh, thanks for spotting! I'll do a new release with this.

PR #199, opened 2024-02-03
NetworkManager: Add stub ipv{4,6} properties in AddWiFiConnection

### LINKED ISSUE #198: Missing ivp4/ipv6 sections in `AddWiFiConnection`
Thanks again for this project!

I am using `AddWiFiConnection`, but when I retrieve the connection settings with [GetSettings](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.Settings.Connection.html#gdbus-method-org-freedesktop-NetworkManager-Settings-Connection.GetSettings) , the sections for `ipv4` and `ipv6` are missing. If I am not mistaken, the settings that are created by the mock function have been simplified on purpose.
Is it possible to add them (even with empty values, to keep retro-compatibility), so that calling `GetSettings` returns output data that looks real? 

I think that an alternative would be to update the settings afterward by using `ConnectionUpdate`. Or is there a better solution?

martinpitt:
Fixes #198

@souliane WDYT?

### DIFF
--- a/dbusmock/templates/networkmanager.py
+++ b/dbusmock/templates/networkmanager.py
@@ -503,7 +503,7 @@ def AddWiFiConnection(self, dev_path, connection_name, ssid_name, _key_mgmt):
 
     The SSID must match one of the previously created access points.
 
-    Please note that this does not set any global properties.
+    Please note that this only sets minimal global properties in ipv{4,6}, no others.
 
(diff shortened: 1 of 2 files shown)


martinpitt:
@pete-woods @thom311 Does that look sensible to you?

PR #200, opened 2024-02-23
power_profiles_daemon: Move to org.freedesktop.UPower.PowerProfiles

martinpitt:
Version 0.20 moved to D-Bus name `org.freedesktop.UPower.PowerProfiles`, and corresponding changes to the object path/interface: https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/releases/0.20

The real daemon also still listens to the old name, but we can't do that in dbusmock: BUS_NAME and friends cannot be parameterized, as their loading and instantiation happens before calling the `load()` function.

`powerprofilesctl` now only looks at the new name, so we need to adjust the template. The new version is in Fedora 39 to rawhide and also in Debian testing, so for ongoing development it's more useful to present the current API.

---

See [recent test failures](https://github.com/martinpitt/python-dbusmock/actions/runs/8014608956)

### DIFF
--- a/dbusmock/templates/power_profiles_daemon.py
+++ b/dbusmock/templates/power_profiles_daemon.py
@@ -15,14 +15,14 @@ This provides only the non-deprecated D-Bus API as of version 0.9.
 __author__ = "Bastien Nocera"
 __copyright__ = """
 (c) 2021, Red Hat Inc.
-(c) 2017 - 2022 Martin Pitt <martin@piware.de>
+(c) 2017 - 2024 Martin Pitt <martin@piware.de>
 """
 
 import dbus
 
-BUS_NAME = "net.hadess.PowerProfiles"
-MAIN_OBJ = "/net/hadess/PowerProfiles"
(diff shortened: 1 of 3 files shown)


martinpitt:
@3v1n0 @hadess FYI -- I'll do a release and Fedora/Debian upload afterwards.

3v1n0:
Well... Apps and libraries are all using the old name though, so not sure if that's too early or she should duplicate it instead?

martinpitt:
@3v1n0 not sure, you tell me -- we could of course copy the whole template as power_profiles_daemon_upower.py or so?

3v1n0:
It seems safer to me, yes.

Maybe using prefixed name though: `upower_power_profiles_daemon.py`?

So we have both and can migrate apps using it without breaking them.

martinpitt:
In retrospect this was a silly move, sorry. I'll retract today's releases and work on a better solution.

martinpitt:
@3v1n0 See PR #201 -- I'd appreciate if you could have a look. Thanks!

PR #201, opened 2024-02-23
Add upower_power_profiles_daemon template

martinpitt:
This works like the existing power_profiles_daemon template, but for the
new D-Bus name/object path introduced in
https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/releases/0.20

Enable integration test on Ubuntu, which has version 0.13 in 22.04 LTS.


martinpitt:
@3v1n0 @hadess FYI

superm1:
I might have broken your plan for it with the proposed change to fix your bug report https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/issues/139.  Can you look at that please?

martinpitt:
@superm1 I saw -- that'll require some adjustment to the parsing indeed. I'll work on a fix ASAP, thanks for the warning.

PR #205, opened 2024-05-08
Add initial ModemManager mock

agx:
We mock the daemon and allow to add a modem (which also adds a SIM). This is enough for `mmcli -m any` and `mmcli -i any` to work and for shells like [phosh](https://phosh.mobi/) to display mobile broadband information.

```
$ mmcli -m any 
  -----------------------------
  General  |              path: /org/freedesktop/ModemManager1/Modems/8
  -----------------------------
  Hardware |             model: E1750
           | firmware revision: 11.126.08.01.00
  -----------------------------
  Status   |             state: enabled
           |       power state: on
           |       access tech: lte
           |    signal quality: 70% (recent)
  -----------------------------
  Modes    |         supported: allowed: 4g; preferred: 4g
           |                    allowed: 2g, 3g; preferred: 3g
           |           current: allowed: 4g; preferred: 4g
  -----------------------------
  3GPP     |              imei: doesnotmatter
           |     operator name: TheOperator
           |      registration: idle
  -----------------------------
  SIM      |  primary sim path: /org/freedesktop/ModemManager1/SIM/2
```

```
$ mmcli -i  any 
  --------------------
  General    |   path: /org/freedesktop/ModemManager1/SIM/2
  --------------------
  Properties | active: yes
             |   imsi: doesnotmatter

```

I intend to expand on this at some point but figured I'll put this out here as it might be useful to others and to figure out the minimal viable thing to land initial support. So feedback would be welcome.


tests/test_modemmanager.py
```diff
@@ -0,0 +1,159 @@
...
+        self.start_mmcli(["-m", "any"])
+
+        self.assertOutputContains(["error: couldn't find modem"])
+
+    def test_modem(self):
+        (modem_path, sim_path) = self.p_obj.AddSimpleModem()
+        self.start_mmcli(["-m", "any"])
+        self.assertOutputEquals(
+            [
+                "  -----------------------------",
```
martinpitt, line 120: How stable is this CLI? Feels a bit amibitious to pin it down to every piece of ASCII art and formatting. It feels a bit more robust to check individual properties with `self.assertIn()` or possibly `assertRegex()`.

But not a veto, this can be relaxed if/when it breaks.

martinpitt (commented):
@agx I don't have much to comment really -- It by and large seems fine (static code tests like pylint and ruff have some complaints, though, see the test failure).

However, I'd suggest to field-test this for a while in your project. upstream templates only really make sense once they are being used by at least two independent projects. If it's better/convenient for you to have upstream, I'm fine with that, though.

Thanks!

tests/test_modemmanager.py
```diff
@@ -0,0 +1,158 @@
...
+class TestModemManagerModemMmcli(TestModemManagerMmcliBase):
+    """main ModemManager interface tests using mmcli"""
+
+    def test_no_modems(self):
+        self.start_mmcli(["-m", "any"])
+
+        self.assertOutputContains(["error: couldn't find modem"])
+
+    def test_modem(self):
+        _ = self.p_obj.AddSimpleModem()
```
martinpitt, line 115: Why the `_` assignment? does anything care about just ignoring the return value?

agx: Merely left that to indicate the method return s.th. potentially useful as (at least I) often use tests to get a code example. Dropped though.

martinpitt:
Thanks! All green. This is still "draft", but I'll let you decide if/when you want this.

agx:
Followed your advice regarding a 2nd application and added a minimal voice interface so gnome-calls can startup and thinks it has a modem (which then is already useful to enable some UI bits like testing the dialpad). Placing calls wouldn't work yet as we don't handle the methods.

I've simplified the test case a bit as we don't need to monitor running mmcli yet (that will only be needed once we support `--monitor-*`) and made sure the new tests don't compare output verbatim but rather match on lines.

I didn't redo the original `mmcli -m any` `mmcli -i any` tests as of yet as they're a nice example of what do expect. But I understand your concerns and am happy to fix that up should mm output change/not be consistent across distros or if I add more to those interfaces (we could e.g. match the first handful of lines and do the rest via regex).

martinpitt:
black and pylint [have some opinion](https://github.com/martinpitt/python-dbusmock/actions/runs/9208138640/job/25363099696?pr=205#step:3:803) here -- want me to fix it up, or want to do that yourself?

agx:
> black and pylint [have some opinion](https://github.com/martinpitt/python-dbusmock/actions/runs/9208138640/job/25363099696?pr=205#step:3:803) here -- want me to fix it up, or want to do that yourself?

Should be fixed now.

agx:
> Sorry, [some more](https://github.com/martinpitt/python-dbusmock/actions/runs/9224708042/job/25381170264?pr=205#step:3:803). If you have trouble with running black/pylint, I'm happy to fix it up myself, quite simple now.

Sorry for the extra trouble. I didn't run those in the right order (and for some reason the workflow doesn't trigger in my repo for this project). I hope it's fixed now (black, ruff and pylint all passed here) - if not I'd be happy about a fixup.

PR #208, opened 2024-06-21
bluez5: Implement Advertising APIs

### LINKED ISSUE #20: Extend Bluez 5 templates to support advertising API
Bluez are adding more dbus API's.

The advertising api is documented at:
http://git.kernel.org/cgit/bluetooth/bluez.git/tree/doc/advertising-api.txt

fkleon:
Basic implementation of the `org.bluez.LEAdvertisingManager1` and `org.bluez.AdvertisementMonitorManager1` interfaces on mock adapters.

Fixes #20

### DIFF
--- a/dbusmock/templates/bluez5.py
+++ b/dbusmock/templates/bluez5.py
@@ -38,8 +38,14 @@ MEDIA_IFACE = "org.bluez.Media1"
 NETWORK_SERVER_IFACE = "org.bluez.Network1"
 DEVICE_IFACE = "org.bluez.Device1"
 
+LE_ADVERTISING_MANAGER_IFACE = "org.bluez.LEAdvertisingManager1"
+LE_ADVERTISEMENT_IFACE = "org.bluez.LEAdvertisement1"
+
 # The device class of some arbitrary Android phone.
 MOCK_PHONE_CLASS = 5898764
+
+# Maximum number of BLE advertisements supported per adapter.
+MAX_ADVERTISEMENT_INSTANCES = 5
 
 
 @dbus.service.method(AGENT_MANAGER_IFACE, in_signature="os", out_signature="")
@@ -107,6 +113,8 @@ def load(mock, _parameters):
     bluez.capabilities = {}
     bluez.default_agent = None
 
+    # whether to expose the LEAdvertisingManager1 interface on adapters (BLE advertising)
+    bluez.enable_advertise_api = _parameters.get("enable_advertise_api", True)
 
 @dbus.service.method(ADAPTER_IFACE, in_signature="o", out_signature="")
 def RemoveDevice(adapter, path):
@@ -221,6 +229,7 @@ def AddAdapter(self, device_name, system_name):
         "Class": dbus.UInt32(268, variant_level=1),  # Computer, Laptop
         "DiscoverableTimeout": dbus.UInt32(180, variant_level=1),
         "PairableTimeout": dbus.UInt32(0, variant_level=1),
+        "Roles": dbus.Array(["central", "peripheral"], variant_level=1),
     }
 
     self.AddObject(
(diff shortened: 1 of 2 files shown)


martinpitt (commented):
Thanks! Starting tests. The detailed unit tests are really nice, but can any of these new functions be tested with `bluetoothctl`? Does it have some "monitor" mode? I.e. I'd have more faith in this if we don't just test our own mock API, but convince the real client beyond just the simple capabilities check in `bluetoothctl list`.

martinpitt:
The [centos 9 failure](https://github.com/martinpitt/python-dbusmock/actions/runs/9613257503/job/26647646973?pr=208#step:3:854) needs some adjustment: 
```
self.assertIn("SupportedCapabilities.MinTxPower: 0xffffffde (-34)", out)

But that older version prints that in two separate lines:

> 'SupportedCapabilities Key: MinTxPower', 'SupportedCapabilities Value: -34'

**Comment by fkleon** [comment]
> The detailed unit tests are really nice, but can any of these new functions be tested with `bluetoothctl`? Does it have some "monitor" mode? I.e. I'd have more faith in this if we don't just test our own mock API, but convince the real client beyond just the simple capabilities check in `bluetoothctl list`.

Good point, and totally possible. Should be able to test both advertising and monitoring features via `bluetoothctl`, so will add those tests.

> But that older version prints that in two separate lines:

Thanks, will have a look into this too.

File: dbusmock/templates/bluez5.py [hunk 1]
```diff
@@ -253,6 +259,47 @@ def AddAdapter(self, device_name, system_name):
...
+            "ActiveInstances": dbus.Byte(0, variant_level=1),
+            "SupportedInstances": dbus.Byte(max_advertisement_instances, variant_level=1),
+            "SupportedIncludes": dbus.Array(["tx-power", "appearance", "local-name", "rssi"], variant_level=1),
+            "SupportedSecondaryChannels": dbus.Array(["1M", "2M", "Coded"], variant_level=1),
+            "SupportedCapabilities": dbus.Dictionary(
+                {
+                    "MaxAdvLen": dbus.Byte(251),
+                    "MaxScnRspLen": dbus.Byte(251),
+                    "MinTxPower": dbus.Int16(-34),
+                    "MaxTxPower": dbus.Int16(7),
```
martinpitt, line 286: Do these numbers have any meaning and come from somewhere? (Like a particular device) or just random?

dbusmock/templates/bluez5.py
```diff
@@ -253,6 +259,47 @@ def AddAdapter(self, device_name, system_name):
...
+        adapter.AddMethods(
+            LE_ADVERTISING_MANAGER_IFACE,
+            [
+                ("RegisterAdvertisement", "oa{sv}", "", RegisterAdvertisement),
+                ("UnregisterAdvertisement", "o", "", UnregisterAdvertisement),
+            ],
+        )
+
+        # Track advertisements per adapter
+        adapter.max_advertisement_instances = max_advertisement_instances
```
martinpitt, line 309: `max_advertisement_instances` never gets changed -- do you plan to do this in the future? Otherwise this is just always going to be 5. I.e. it could just as well be a global constant.

dbusmock/templates/bluez5.py
```diff
@@ -652,3 +699,69 @@ def DisconnectDevice(_self, adapter_device_name, device_address):
...
+    """Convenience method to add an Advertisement object
+
+    Returns the new object path.
+    """
+    path = "/org/dbusmock/bluez/advertisement/" + adv_name
+
+    adv_properties = {
+        "Type": dbus.String("broadcast", variant_level=1),
+        "ManufacturerData": dbus.Dictionary(
+            {dbus.Int16(0x0397): dbus.Array([0x00, 0x01], variant_level=2)}, signature="qv", variant_level=1
```
martinpitt, line 779: Again, magic numbers -- this could use a comment where these come from, or what they mean?

tests/test_bluez5.py
```diff
@@ -128,6 +128,24 @@ def test_one_adapter(self):
...
+        self.assertIn("SupportedIncludes: tx-power", out)
+        self.assertIn("SupportedIncludes: appearance", out)
+        self.assertIn("SupportedIncludes: local-name", out)
+        self.assertIn("SupportedSecondaryChannels: 1M", out)
+        self.assertIn("SupportedSecondaryChannels: 2M", out)
+        self.assertIn("SupportedCapabilities.MinTxPower: 0xffffffde (-34)", out)
+        self.assertIn("SupportedCapabilities.MaxTxPower: 0x0007 (7)", out)
+        self.assertIn("SupportedCapabilities.MaxAdvLen: 0xfb (251)", out)
+        self.assertIn("SupportedCapabilities.MaxScnRspLen: 0xfb (251)", out)
+        self.assertIn("SupportedFeatures: CanSetTxPower", out)
```
martinpitt, line 148: This [fails in CentOS 9 ](https://github.com/martinpitt/python-dbusmock/actions/runs/9743502345/job/26898417289?pr=208)

tests/test_bluez5.py
```diff
@@ -128,6 +128,24 @@ def test_one_adapter(self):
...
+        self.assertIn("SupportedFeatures: HardwareOffload", out)
```
martinpitt, line 149: This also isn't in the [centos 9 output](https://github.com/martinpitt/python-dbusmock/actions/runs/9743502345/job/26898417289?pr=208#step:3:862) -- hmm, these are exactly the two you advertise in the mock. It works everywhere else, so this must be related to an older client version. Maybe you have an idea why, but if not, let's just make these conditional?

fkleon: Looks like `SupportedFeatures` was only added to the API with BlueZ 5.57 (CentOS 9 is on 5.56). I'll make this test conditional on the detected BlueZ version as well.

dbusmock/templates/bluez5.py
fkleon: Yes, good point. Was vaguely thinking about making this configurable with a parameter to the template, but then didn't see a compelling use case, so a constant would do just fine.

dbusmock/templates/bluez5.py
fkleon: The TX power values are in dbm and I believe are regulated depending on BT device type. I've grabbed these and the other values them from my bluetooth adapter (Intel AX200). Happy to document that if needed.

fkleon: Added a brief comment about the source of the values.

dbusmock/templates/bluez5.py
fkleon: Made this a constant instead.

dbusmock/templates/bluez5.py
fkleon: Have added some comments and switched to the reserved company identifier `0xFFFF` which is specifically for testing.

tests/test_bluez5.py
fkleon: Hopefully resolved now! :)

martinpitt (approved):
Amazing  work, thank you! Do you want this in a release (I can do one after landing), or is it enough for you to be in  main?

fkleon:
> Amazing work, thank you! Do you want this in a release (I can do one after landing), or is it enough for you to be in main?

Thanks for the quick review! Happy to source this from main until the next release lands, not urgent at all.

PR #213, opened 2024-08-02
tests: Drop outdated doc strings

agx:
Two files tried to documen what is being tested and got it wrong.  Drop that duplicate informtion as it's obvious from the class names.


martinpitt (approved):
Hah, copy pasta. Thanks!

PR #218, opened 2024-09-13
mockobject: Allow adding objects derived from DBusMockObject

swick:
This allows to create objects instanciated from a DBusMockObject subclass in mock templates. That way, implementing dynamic objects becomes easier by using dbus decorators:

```python
    class ColordDevice(mockobject.DBusMockObject):
        @dbus.service.method(DEVICE_IFACE, in_signature='', out_signature='i')
	def Test(self):
		return 42

    self.add_object(ColordDevice, device_path, DEVICE_IFACE, {}, [])
```

### DIFF
--- a/dbusmock/mockobject.py
+++ b/dbusmock/mockobject.py
@@ -199,6 +199,7 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         props: PropsType,
         logfile: Optional[str] = None,
         is_object_manager: bool = False,
+        mock_data: Any = None,
     ) -> None:
         """Create a new DBusMockObject
 
@@ -234,6 +235,7 @@ class DBusMockObject(dbus.service.Object):  # pylint: disable=too-many-instance-
         self.logfile = open(logfile, "wb") if logfile else None  # noqa: SIM115
         self.is_logfile_owner = True
         self.call_log: List[CallLogType] = []
+        self.mock_data = mock_data
 
         if props is None:
             props = {}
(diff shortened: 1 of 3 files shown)


swick:
I want to use this in the mutter test suite: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4022/diffs

dbusmock/mockobject.py
```diff
@@ -322,6 +322,23 @@ def Set(self, interface_name: str, property_name: str, value: Any, *_, **__) ->
             [interface_name, dbus.Dictionary({property_name: value}, signature="sv"), dbus.Array([], signature="s")],
         )
 
+    def add_object(self, mock_class: Type[Self], path: str, interface: str, properties: PropsType, methods: List[MethodType]) -> Self:
```
martinpitt: Having another public method on this class feels a bit confusing. Why not simply add a kwarg to the existing AddObject? If that somehow interferes with the decorator, then can you please make it private? Like `_add_object_with_class`?

martinpitt (changes requested):
Thanks! Sorry for the late review, I missed that during my vacation.

This makes sense, but needs documentation (in the docstrings) and unit tests. Please let me know if you need help with either.

dbusmock/mockobject.py
swick: I did add it as kwargs now but the arguments `mock_class` and `mock_data` are only useful when the function is called directly and not via dbus.

swick:
Tried to add documentation and a unit test. Let me know if it needs adjusting.

martinpitt:
@swick thanks! I'm AFK until next Wednesday, will re-review then. But at a quick glance this looks great!

martinpitt:
There's some pylint/ruff/black failures, see [test failure](https://github.com/martinpitt/python-dbusmock/actions/runs/11162107009/job/31046425737?pr=218#step:4:847). If you don't feel like fixing these  (installing these tools and running the tests), I'm happy to do that, too.

swick:
I think I managed to fix it. If something is still wrong, feel free to take over.

martinpitt:
Thanks @swick ! I rebased, fixed a trivial black error, and some docstring formatting.

martinpitt (approved):

swick:
Thank you!

swick:
Hey @martinpitt, we would like to depend on this change for mutter. Any chance we can get a release soon?

martinpitt:
@swick Oh sorry, I thought this was already included in 0.32.2, but that was a few days earlier. No problem!

New release: https://github.com/martinpitt/python-dbusmock/releases/tag/0.33.0

Packaging for Fedora/Debian/Ubuntu now.

PR #221, opened 2024-10-15
tests: fix bluez test when no /etc/os-release

alyssais:
This file doesn't exist in the Nix build sandbox.

Fixes: 812dc61 ("tests: Skip bluez test_advertise on RHEL 10")


martinpitt (approved):
Thank you!

PR #222, opened 2024-10-15
tests: fix test_monitor on musl

alyssais:
[It is probably a Glibc bug that sscanf as used by bluez accepts 0x01 for this value.][1]  For portability, use the unprefixed hex value. Based on bluez's own [documentation][2], I don't think the 0x is required.

[1]: https://www.openwall.com/lists/libc-coord/2024/10/15/2
[2]: https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/client/bluetoothctl-monitor.rst?id=3d9900eb754d277ef6abf21bed827cdc9e79ddb4

### DIFF
--- a/tests/test_bluez5.py
+++ b/tests/test_bluez5.py
@@ -493,7 +493,7 @@ class TestBlueZ5(dbusmock.DBusTestCase):
         adapter = self.dbus_con.get_object("org.bluez", path)
 
         # When an advertisement monitor is configured via bluetoothctl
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Ugh, starting the day with reading the (thankfully long past) exclamations from Ulrich Drepper.. But aside from this, thanks for that insight into the sscanf() mess.

Thank you!

PR #227, opened 2024-12-18
modemmanager: Allow to add Cell broadcast messages

agx:
Doing userspace development for cell broadcast messages can be tricky as they're very rare in the wild. If one doesn't mock ModemManager one needs to have a small GSM cell for testing.
    
Hence add a `AddCbm()` message to add cell broadcast messages to the MM mock.
    
```
 mm_server.obj.AddCbm(2, 4383, "This is a test)
```
    
Messages can be deleted via MMs API:
    
```
busctl call --system org.freedesktop.ModemManager1 /org/freedesktop/ModemManager1/Modems/8 org.freedesktop.ModemManager1.Modem.CellBroadcast Delete o /org/freedesktop/ModemManager1/Cbm/5
```
    
and listed via MMs API:
    
```
busctl call --system org.freedesktop.ModemManager1 /org/freedesktop/ModemManager1/Modems/8 org.freedesktop.ModemManager1.Modem.CellBroadcast List
```

This functionality is not yet in a released MM version (but merged into main). I've tested that MMs `mmcbmmonitor` handles it, [phosh](https://gitlab.gnome.org/World/Phosh/phosh) can use it to test its CBM display and we intend to use it to implement CBM support in [Chatty](https://gitlab.gnome.org/World/Chatty).

As a preparation we improve the general MM implementation. We so far used the default ObjectManager but that would be too much as e.g. SIM (or CBM) shouldn't be announced via that interface. Hence we implement `GetManagedObjects` ourself to only handle modems.

### DIFF
--- a/dbusmock/templates/modemmanager.py
+++ b/dbusmock/templates/modemmanager.py
@@ -25,8 +25,11 @@ SYSTEM_BUS = True
 IS_OBJECT_MANAGER = False
 MODEM_IFACE = "org.freedesktop.ModemManager1.Modem"
 MODEM_3GPP_IFACE = "org.freedesktop.ModemManager1.Modem.Modem3gpp"
+MODEM_CELL_BROADCAST_IFACE = "org.freedesktop.ModemManager1.Modem.CellBroadcast"
 MODEM_VOICE_IFACE = "org.freedesktop.ModemManager1.Modem.Voice"
 SIM_IFACE = "org.freedesktop.ModemManager1.Sim"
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks @agx ! This looks sensible from an "engineer who knows nothing about MM" perspective, but you are the SME :grin: 

I'll do a release this week still, as I want to test the recent changes to the release workflow.

martinpitt:
@agx: Released in https://github.com/martinpitt/python-dbusmock/releases/tag/0.34.0

agx:
Great news, thank you!

PR #231, opened 2025-02-21
tests: Relax libnotify expected format for libnotify 0.8.4

martinpitt:


martinpitt:
@3v1n0 : I just noticed that python-dbusmock's regular upstream unit tests started to fail for Debian unstable, and https://tracker.debian.org/pkg/libnotify has the same problem. This PR fixes it. I'll do a new upstream release and upload it to Debian now.

PR #234, opened 2025-05-22
modemmanager: Add operator ID and allow to set channels

agx:
The channel setting API is not yet in a released version of ModemManager so I didn't add a `mmcli` test.

### DIFF
--- a/dbusmock/templates/modemmanager.py
+++ b/dbusmock/templates/modemmanager.py
@@ -189,6 +189,7 @@ def AddSimpleModem(self):
     modem_3gpp_props = {
         "Imei": dbus.String("doesnotmatter"),
         "OperatorName": dbus.String("TheOperator"),
+        "OperatorCode": dbus.String("00101"),
         "Pco": dbus.Array([], signature="(ubay)"),
     }
     modem = mockobject.objects[modem_path]
--- a/dbusmock/templates/modemmanager.py
+++ b/dbusmock/templates/modemmanager.py
@@ -151,6 +151,17 @@ def deleteCbm(self, cbm_path):
     )
 
 
+def setChannels(_self, channels):
+    modem_obj = mockobject.objects[SIMPLE_MODEM_PATH]
+
+    modem_obj.UpdateProperties(
+        MODEM_CELL_BROADCAST_IFACE,
+        {
+            "Channels": dbus.Array(channels),
+        },
+    )
+
+
 @dbus.service.method(MOCK_IFACE, in_signature="", out_signature="ss")
 def AddSimpleModem(self):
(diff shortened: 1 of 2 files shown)


agx:
The debian and ubuntu test fail to install packages and the fedora tests fails on ruff in unchanged files (https://github.com/martinpitt/python-dbusmock/actions/runs/15196820126/job/42742834043?pr=234#step:4:825) .

dbusmock/templates/modemmanager.py
```diff
@@ -196,11 +207,13 @@ def AddSimpleModem(self):
     modem.AddProperties(MODEM_3GPP_IFACE, modem_3gpp_props)
 
     modem_cell_broadcast_props = {
+        "Channels": dbus.Array([], signature="(uu)"),
         "CellBroadcasts": dbus.Array([], signature="o"),
     }
     modem_cell_broadcast_methods = [
         ("List", "", "ao", listCbm),
         ("Delete", "o", "", deleteCbm),
+        ("SetChannels", "a(uu)", "", setChannels),
```
martinpitt, line 216: OK, these match https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/commit/2878833f5c83fe112f579865a6d318ce8413f452

Let's see if the next MM upstream release will break the tests due to changing the output format. I actually wish it did, then we'd have a good reminder to check this new option :wink:

martinpitt (approved):
This looks good, thanks! A bit sad that this is hard to test, is there perhaps some COPR or so with the new version? I'm happy to add an integration test that pulls in some "git main builds" packages.

martinpitt:
I  pushed a clean rebase to your branch.

dbusmock/templates/modemmanager.py
agx: I think it won't. We need to additionally run `mmcli -m any --cell-broadcast-status` for that so it won't break out of a sudden.

agx:
> This looks good, thanks! A bit sad that this is hard to test, is there perhaps some COPR or so with the new version? I'm happy to add an integration test that pulls in some "git main builds" packages.

Upstream MM tags dev snapshots every now and then but that didn't happen yet.

martinpitt:
@agx Released as https://github.com/martinpitt/python-dbusmock/releases/tag/0.35.0 . Uploading to Fedora and Debian experimental now (due to the release freeze)

agx:
> @agx Released as https://github.com/martinpitt/python-dbusmock/releases/tag/0.35.0 . Uploading to Fedora and Debian experimental now (due to the release freeze)

That was quick, thanks a lot!

PR #239, opened 2025-06-04
mockobject: Fix _wrap_in_dbus_variant for Struct and Dict types

swick:
Updating a Struct or Dict property fails because they have no conjugate method. We don't need to wrap them, so they really behave like Array, so let's just treat them in the same way.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -343,6 +343,32 @@ assert args[2] == 5
         self.assertEqual(self.dbus_props.GetAll("org.freedesktop.Test.Other"), {"color": "yellow"})
         self.assertEqual(self.dbus_props.Get("org.freedesktop.Test.Other", "color"), "yellow")
 
+        # add properties with complex types
+        self.dbus_mock.AddProperty("org.freedesktop.Test.Main", "intarray", dbus.Array([1, 2]))
+        self.dbus_mock.AddProperty("org.freedesktop.Test.Main", "vardict", dbus.Dictionary({"a": 1}))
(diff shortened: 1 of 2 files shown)


martinpitt:
Thanks for catching this! This should get a test in tests/test_api.py. That code is only used in `UpdateProperties()`, so it should be added to `test_properties()` -- add an array property and change it. Do you want to do that, or want me to?

swick:
Added tests to `test_properties()`

tests/test_api.py
```diff
@@ -357,28 +383,26 @@ def catch(*args, **kwargs):
...
 
         GLib.timeout_add(3000, ml.quit)
         ml.run()
 
         match.remove()
 
-        self.assertEqual(self.dbus_props.GetAll("org.freedesktop.Test.Main"), {"version": 5, "connected": False})
-        self.assertEqual(changed_props, [{"version": 5, "connected": False}])
+        for i, _ in enumerate(changes):
+            self.assertEqual(changed_props[i], changes[i])
```
martinpitt, line 405: Could be a bit more elegant with `for i, change in ..` and comparing against `change`, but that's total bikeshedding.

martinpitt (dismissed):
This looks great, thank you! Waiting for tests.

tests/test_api.py
martinpitt: Lol, that's exactly what [pylint complains about](https://github.com/martinpitt/python-dbusmock/actions/runs/15463734765/job/43530844659?pr=239#step:4:848). So can you please do this still?

martinpitt (approved):
Danke sehr!

